PR #34, opened 2026-06-04
Fix MediaIngredientMech linker for CHEBI-keyed MIM schema

realmarcin:
## Why
MediaIngredientMech migrated from minting `MediaIngredientMech:NNNNNN` ids to keying its curated ingredients by ontology identifier (predominantly CHEBI). The CultureMech loader still indexed by `id` expecting the legacy format, so it loaded **0 entries** and matched nothing — a silent, **repo-wide** breakage of the MIM enrichment step.

## What
- **loader**: index by `identifier` / `ontology_mapping.ontology_id`. Indexes go 0 → **1726 CHEBI / 1866 name / 4643 synonym**.
- **linker**: resolve each match to its CHEBI grounding and write a new `mediaingredientmech_chebi_term`; skip matches whose MIM entity is not CHEBI-keyed (~13%: FOODON/ENVO/NCIT/…).
- **schema**: add `MediaIngredientMechChebiTerm` class + the `mediaingredientmech_chebi_term` slot on ingredient, solution, and solution-component. Legacy `mediaingredientmech_term` retained + documented for the pre-migration corpus.
- regenerate `culturemech_dataclasses.py` (the 21 dataclass-dependent tests pass).

## Not included
`culturemech.schema.json` is **independently stale** (missing pre-existing `AtmosphereEnum`/`ChebiTerm`/etc.) and is intentionally left untouched here to avoid ~2k lines of unrelated churn. It deserves its own regeneration PR.


src/culturemech/enrich/mediaingredientmech_loader.py
```diff
@@ -91,15 +91,29 @@ def _build_indexes(self):
...
+                ontology_mapping.get('ontology_id')
+                or ingredient.get('ontology_id')
+                or ingredient.get('chebi_id')
+            )
+            if not chebi_id and str(mim_id).startswith('CHEBI:'):
+                chebi_id = mim_id
             if chebi_id:
                 # Normalize CHEBI ID format
-                if not chebi_id.startswith('CHEBI:'):
+                if not str(chebi_id).startswith('CHEBI:'):
```
realmarcin: RISK: `chebi_id` here can hold a non-CHEBI CURIE. In the migrated schema `ontology_mapping.ontology_id` is a full CURIE of any ontology (FOODON/ENVO/NCIT, the ~13% the PR calls out). For those, `str(chebi_id).startswith('CHEBI:')` is False, so this branch prepends `CHEBI:` and produces malformed keys like `CHEBI:FOODON:03304010`, which then get inserted into `by_chebi`. This pollutes the strictly-CHEBI index and inflates the reported CHEBI count (the prepend was only meant to normalize legacy bare-numeric ids). Fix: only normalize bare numerics and only index real CHEBI ids, e.g. `chebi_id = str(chebi_id); if chebi_id.isdigit(): chebi_id = f'CHEBI:{chebi_id}'` then `if chebi_id.startswith('CHEBI:'): self.by_chebi[chebi_id] = ingredient`.

src/culturemech/enrich/mediaingredientmech_linker.py
```diff
@@ -59,14 +61,33 @@ def enrich_ingredient(self, ingredient: Dict[str, Any]) -> bool:
...
+            chebi_id = ontology_mapping.get('ontology_id')
+            if not (chebi_id and str(chebi_id).startswith('CHEBI:')):
+                ident = match.get('identifier') or match.get('id') or ''
+                chebi_id = ident if str(ident).startswith('CHEBI:') else None
+
+            if not chebi_id:
+                # Matched a MIM entity that is not CHEBI-keyed -> no strictly
+                # CHEBI linkage to record.
+                self.stats['no_match'] += 1
+                self.stats.setdefault('matched_non_chebi', 0)
```
realmarcin: NIT: a name/synonym/fuzzy hit on a non-CHEBI MIM entity is counted as `no_match` (plus `matched_non_chebi`) but is not appended to `self.unmatched_ingredients`, unlike the genuine no-match path below. The unmatched report will therefore omit these ingredients even though they were effectively unmatched for CHEBI linkage. Consider appending to `unmatched_ingredients` here too (or tracking them in a dedicated list) so the report stays complete.

realmarcin (commented):
This PR fixes the MediaIngredientMech (MIM) enrichment loader/linker for MIM's migration from minted `MediaIngredientMech:NNNNNN` ids to a CHEBI-keyed schema, and adds a `MediaIngredientMechChebiTerm` class + `mediaingredientmech_chebi_term` slot (with regenerated dataclasses) while retaining the legacy term for the existing corpus. The schema/dataclass changes are consistent and well-documented, and the dual-field backward-compat logic in loader/linker is sound. Two findings below; one is a real index-correctness issue.

- NIT (mediaingredientmech_linker.py, enrich_solution ~line 122, pre-existing, not in this diff): the solution-level skip still checks only `mediaingredientmech_term` and not the new `mediaingredientmech_chebi_term`, and it increments `already_linked` without returning (continues processing composition). Now that `enrich_ingredient` checks both fields, consider aligning `enrich_solution` for consistency.

src/culturemech/enrich/mediaingredientmech_linker.py
```diff
@@ -59,14 +61,41 @@ def enrich_ingredient(self, ingredient: Dict[str, Any]) -> bool:
...
+                # Matched a MIM entity that is not CHEBI-keyed -> no strictly
+                # CHEBI linkage to record. Surface it in the unmatched report
+                # so name/synonym/fuzzy hits on non-CHEBI MIM entities are
+                # visible, not silently dropped.
+                self.stats['no_match'] += 1
+                self.stats.setdefault('matched_non_chebi', 0)
+                self.stats['matched_non_chebi'] += 1
+                self.unmatched_ingredients.append({
+                    'name': name,
+                    'chebi_id': chebi_id or 'N/A',
```
realmarcin, line 102: BUG: `chebi_id` was reassigned at lines 75-78 to the matched MIM entity's CHEBI id and is `None` inside this `if not chebi_id:` branch, so `chebi_id or 'N/A'` always records `'N/A'` here — the ingredient's own query CHEBI id (computed at line 58) is lost in the unmatched report. Fix: keep the ingredient's query id in a separate variable (e.g. `query_chebi_id = ingredient['term'].get('id')` at line 58) and reference that here, e.g. `'chebi_id': query_chebi_id or 'N/A'`, rather than the overwritten `chebi_id`.

src/culturemech/enrich/mediaingredientmech_linker.py
```diff
@@ -98,7 +127,9 @@ def enrich_solution(self, solution: Dict[str, Any]) -> int:
         matched_count = 0
 
         # Skip if already has MediaIngredientMech link at solution level
-        if 'mediaingredientmech_term' in solution:
+        # (legacy id-based or current CHEBI-keyed), matching enrich_ingredient.
+        if ('mediaingredientmech_term' in solution
+                or 'mediaingredientmech_chebi_term' in solution):
```
realmarcin: NIT: this block increments `self.stats['already_linked']` when the solution already has a (legacy or CHEBI) link but does NOT `return` — it falls through and still enriches every `composition` ingredient. The comment says "matching enrich_ingredient," but `enrich_ingredient` returns immediately after the same check, so behavior is not actually matched and `already_linked` is incremented even though nothing is skipped. Either `return matched_count` here (if a solution-level link should suppress composition enrichment) or drop the misleading stat increment + comment (if per-ingredient enrichment is intended regardless).

cache/chebi/terms.csv
```diff
@@ -0,0 +1,15 @@
+curie,label,retrieved_at
```
realmarcin: NIT: this is a runtime CHEBI label cache being committed as a tracked file. `cache/` is not gitignored and is otherwise untracked, so this looks unintentional and will cause churn/staleness. (I verified all 14 labels against the local CHEBI ontology — they are all correct, so no data issue.) Suggest adding `cache/` to `.gitignore` and dropping this file from the PR.

realmarcin (commented):
## Summary

This PR fixes the MediaIngredientMech (MIM) enrichment loader/linker for MIM's migration from minted `MediaIngredientMech:NNNNNN` ids to a CHEBI-keyed schema, indexing by `identifier`/`ontology_mapping.ontology_id` (0 → 1726/1866/4643 index entries), adds a new `mediaingredientmech_chebi_term` schema class+slots, and regenerates the dataclasses. The core fix is sound: `_build_indexes` correctly promotes only bare-numeric CHEBI values to CURIEs (guarding against mangling `FOODON:`/`ENVO:` ids), the CHEBI-only linkage guard is correct, and the legacy `mediaingredientmech_term` path is preserved. I verified all 14 CHEBI labels in the new `cache/chebi/terms.csv` against the local CHEBI ontology — every label is correct, no fabricated/mismatched terms.

Findings below. Nothing blocks the core fix; one report-only logic bug and two NITs.

### BUG
- `mediaingredientmech_linker.py:90` — variable-shadowing drops the ingredient's query CHEBI id from the unmatched report (always logs `N/A`).

### NIT
- `mediaingredientmech_linker.py:132` — `enrich_solution` increments `already_linked` without returning; comment claims parity with `enrich_ingredient` (which does return).
- `cache/chebi/terms.csv` — runtime label cache committed as a tracked file (not gitignored); likely unintentional churn.

- RISK `mediaingredientmech_loader.py:~195` (`find_match`, unchanged by this PR but now inconsistent): the CHEBI-id normalization still unconditionally prepends `CHEBI:` to any non-CHEBI input (`if not chebi_id.startswith('CHEBI:'): chebi_id = f"CHEBI:{chebi_id}"`), whereas the fixed `_build_indexes` now only promotes bare-numeric values. Impact is low because `by_chebi` only holds genuine CHEBI keys (a mangled `CHEBI:GO:...` simply misses), but the two code paths should use the same normalization rule (promote only when `chebi_id.isdigit()`) to avoid confusion.
- Note (no change needed): in the non-CHEBI match branch the matched MIM entity is counted under both `no_match` and the new `matched_non_chebi` stat and appended to `unmatched_ingredients`. This appears intentional per the inline comment, but `no_match` now conflates true misses with CHEBI-less matches — worth a docstring note on the returned stats if any downstream consumer treats `no_match` as "no MIM entity found".

src/culturemech/enrich/mediaingredientmech_linker.py
```diff
@@ -59,14 +61,42 @@ def enrich_ingredient(self, ingredient: Dict[str, Any]) -> bool:
...
+                self.stats['matched_non_chebi'] += 1
+                self.unmatched_ingredients.append({
+                    'name': name,
+                    'chebi_id': chebi_id or 'N/A',
+                    'matched_non_chebi': match.get('identifier')
+                    or match.get('id') or 'N/A',
+                })
+                return False
+
+            ingredient['mediaingredientmech_chebi_term'] = {
```
realmarcin, line 123: RISK: this writes `id` = the MIM entity's CHEBI grounding but `label` = MIM `preferred_term` (set above). The MIM preferred_term is not guaranteed to equal that CHEBI term's rdfs:label, so the emitted term can carry an id/label pair that disagrees. Separately, because `find_match` returns name/synonym/fuzzy (>=0.95) hits in addition to exact CHEBI-id hits, a fuzzy name match will assert a CHEBI id that may differ from the ingredient's own `term`/`chebi_term` grounding. Consider either (a) only writing the CHEBI id when `match_method == 'chebi_id'`, or (b) recording `match_method` on the emitted term so downstream consumers don't conflate a fuzzy MIM linkage with a curated CHEBI grounding.

src/culturemech/enrich/mediaingredientmech_linker.py
```diff
@@ -59,14 +61,42 @@ def enrich_ingredient(self, ingredient: Dict[str, Any]) -> bool:
...
+                ident = match.get('identifier') or match.get('id') or ''
+                mim_chebi_id = ident if str(ident).startswith('CHEBI:') else None
+
+            if not mim_chebi_id:
+                # Matched a MIM entity that is not CHEBI-keyed -> no strictly
+                # CHEBI linkage to record. Surface it in the unmatched report
+                # so name/synonym/fuzzy hits on non-CHEBI MIM entities are
+                # visible, not silently dropped. Report the ingredient's own
+                # query CHEBI id (not the MIM match, which has no CHEBI id).
+                self.stats['no_match'] += 1
```
realmarcin: NIT: a match WAS found here (just not CHEBI-keyed), yet this increments `no_match`. The `No match found` summary line then conflates true no-matches with matched-but-non-CHEBI MIM entities. Consider reporting `matched_non_chebi` as its own summary line and not adding it into `no_match`, so the breakdown stays interpretable.

src/culturemech/enrich/mediaingredientmech_linker.py
```diff
@@ -98,7 +128,9 @@ def enrich_solution(self, solution: Dict[str, Any]) -> int:
         matched_count = 0
 
         # Skip if already has MediaIngredientMech link at solution level
-        if 'mediaingredientmech_term' in solution:
+        # (legacy id-based or current CHEBI-keyed), matching enrich_ingredient.
+        if ('mediaingredientmech_term' in solution
+                or 'mediaingredientmech_chebi_term' in solution):
```
realmarcin: NIT (pre-existing, surfaced by this edit): when a solution already carries a MIM link this increments `already_linked` but does not `return`, so the composition ingredients are still iterated and enriched below. If that's intentional (solution-level link != ingredient-level), the `already_linked` increment is misleading; if not, add a `return 0`.

src/culturemech/enrich/mediaingredientmech_loader.py
```diff
@@ -91,17 +91,37 @@ def _build_indexes(self):
...
+                chebi_id = str(chebi_id).strip()
+                # Normalize CHEBI ID format. Only a BARE NUMERIC value should be
+                # promoted to a CHEBI CURIE — otherwise a non-CHEBI grounding
+                # (FOODON/ENVO/NCIT, e.g. `FOODON:03304010`) would be mangled
+                # into a malformed key like `CHEBI:FOODON:03304010`.
+                if chebi_id.isdigit():
                     chebi_id = f"CHEBI:{chebi_id}"
-                self.by_chebi[chebi_id] = ingredient
+                # Only genuine CHEBI ids belong in the CHEBI-keyed index.
+                if chebi_id.startswith('CHEBI:'):
```
realmarcin, line 127: NIT: `startswith('CHEBI:')` is case-sensitive. A grounding stored as `chebi:NNN` (lowercase prefix) would pass the earlier `isdigit()` check as False and then be silently dropped from `by_chebi`. If MIM data is guaranteed uppercase this is fine; otherwise normalize the prefix case before this check.

realmarcin (commented):
Solid, well-documented fix for the real repo-wide breakage (MIM CHEBI-keyed migration left the loader indexing by the dead `id` field, loading 0 entries). The loader's bare-numeric-only CHEBI promotion correctly avoids mangling FOODON/ENVO/NCIT ids into `CHEBI:FOODON:...`, and the schema + regenerated dataclasses are internally consistent. The notes below are about emitted-data integrity and stats reporting, not the core fix. Note: this diff asserts no concrete CHEBI/GO/NCBITaxon id↔label pairs (only generic `^CHEBI:\d+$` patterns and an illustrative `FOODON:03304010` comment), so there was nothing to verify against local OAK.

## Severity-grouped findings

**RISK**
- `mediaingredientmech_linker.py:97` — emitted `mediaingredientmech_chebi_term` pairs a CHEBI `id` (MIM grounding) with a `label` taken from MIM `preferred_term`, which is not guaranteed to equal that CHEBI term's rdfs:label; downstream consumers reading this as a CHEBI grounding may see id/label disagreement.
- `mediaingredientmech_linker.py:97` — name/synonym/fuzzy(>=0.95) matches now all write a CHEBI-id-bearing term, so a high-but-imperfect name hit can assert a CHEBI id that differs from the ingredient's own `term`/`chebi_term`.

**NIT**
- `mediaingredientmech_linker.py:86` — non-CHEBI matches increment both `no_match` and `matched_non_chebi`, so the `No match found` summary line is inflated by matched-but-non-CHEBI hits.
- `mediaingredientmech_linker.py:133` — solution-level already-linked branch increments `already_linked` but does not `return`/skip, so composition is still re-processed (pre-existing, but the condition was touched here).
- `mediaingredientmech_loader.py:123` — `startswith('CHEBI:')` is case-sensitive; a lowercase `chebi:NNN` grounding would be silently dropped from `by_chebi`.

None beyond the above; the schema (`culturemech.yaml`) and regenerated `culturemech_dataclasses.py` additions are consistent with each other and with the new `MediaIngredientMechChebiTerm` pattern.

realmarcin:
## Review findings addressed (commit pushed)

Resolved all open threads. Summary of fixes in this PR's latest commit + the two prior follow-up commits:

**Loader** (`mediaingredientmech_loader.py`)
- RISK (line ~116, non-CHEBI CURIE mangling): fixed — only bare-numeric values are promoted to `CHEBI:`, and only genuine `CHEBI:` keys are indexed into `by_chebi` (no more `CHEBI:FOODON:…`). The loader also now reads flat `ontology_id`/`chebi_id` fallbacks.
- NIT (line ~123, case-sensitive prefix): fixed — a lowercase/mixed-case `chebi:` prefix is now normalized to `CHEBI:` before indexing.

**Linker** (`mediaingredientmech_linker.py`)
- BUG (line ~91, query-CHEBI-id lost to `N/A`): fixed — the ingredient's own query id is kept in `chebi_id` (the MIM match id lives in the separate `mim_chebi_id`), so the unmatched report records the real query id.
- RISK (line ~97, id/label disagreement + fuzzy id assertion): fixed — fuzzy (>=0.95) matches no longer assert a CHEBI link (the resolved CHEBI id can differ from the ingredient's own grounding); they're counted in `fuzzy_skipped` and surfaced in the report. id/label semantics (MIM `preferred_term` as label) are documented on the `MediaIngredientMechChebiTerm` schema class.
- NIT (line ~86, `no_match` conflation): fixed — matched-but-non-CHEBI hits increment a dedicated `matched_non_chebi` stat (and a new summary line) instead of inflating `no_match`.
- NIT (line ~87, non-CHEBI hits omitted from report): fixed — they're now appended to `unmatched_ingredients`.
- NIT (line ~78, flat-field parity): fixed — linker resolves `ontology_mapping.ontology_id` → flat `ontology_id`/`chebi_id` → `identifier`/`id`, matching the loader.
- NIT (line ~133, solution `already_linked` without return): fixed — removed the misleading increment; `enrich_solution` only links composition ingredients and each `enrich_ingredient` handles its own already-linked bookkeeping.

**Cache** (`cache/chebi/terms.csv`)
- NIT: fixed — `cache/` is now gitignored and the runtime label cache is untracked.

Also updated the stale unit tests to assert the migrated `mediaingredientmech_chebi_term` field.

PR #38, opened 2026-06-05
Refresh 120k legacy MediaIngredientMech links to CHEBI keying

realmarcin:
## Why
MediaIngredientMech deprecated its `MediaIngredientMech:NNNNNN` id scheme and re-keyed curated ingredients by CHEBI (current canonical file carries **0** of the old ids). **9,695 CultureMech records** still referenced that dead scheme via `mediaingredientmech_term`.

## What — id-safe refresh-replace
`scripts/migrate_legacy_mim_to_chebi.py` rewrites legacy links to `mediaingredientmech_chebi_term`:

- **CHEBI present (~99.8%)** — the ingredient already carries its own CHEBI `term`; if that CHEBI is in current MIM, replace the legacy field using the ingredient's **own** CHEBI id + label. MIM's `preferred_term` is **not** imported (it diverges ~21%), so no ingredient labels change semantically.
- **CHEBI gap (~203)** — legacy link, no CHEBI term → name/synonym match against current MIM; on a hit, also set the ingredient's `term` (closes the grounding gap) + add the link. (Verified: "Vitamine B12" → CHEBI:176843.)
- **Unmappable (1,509 records)** — CHEBI not in current MIM and no name match → legacy field left intact.

## Result
- **9,671 files changed, 120,029 links refreshed.**
- Re-emitted with PyYAML `safe_dump` (the corpus's original settings) → unchanged fields round-trip **byte-identically**, diffs are per-field. The +56k net insertions are the per-file `curation_history` entries.
- All 9,671 migrated files parse; sampled + corpus-wide checks confirm **0** non-CHEBI ids in the new field.
- Reviewers: review **the script + a few sample diffs + the report**, not all 9,671 files (mechanical, uniform transform).

## Byproducts folded in
- **203-gap fix** — no-CHEBI stragglers grounded via MIM name match.
- **Grounding-audit flag** — `data/import_tracking/reports/mim_grounding_divergences.tsv` (500 CHEBI ids where CultureMech vs MIM labels differ). Mostly benign formatting (`Na-ascorbate`), with a thin tail of real issues (e.g. `CHEBI:37583` phosphate confusion; 2 pyridoxine/pyridoxamine mislabels) for `audit-schema-gaps`.

## Dependency
⚠️ **Stacked on #34** (uses the `MediaIngredientMechChebiTerm` schema field). Base retargets to `main` once #34 merges.


realmarcin (commented):
Mechanical legacy-MIM → CHEBI link refresh. Reviewed the migration script + the three reports + sampled/aggregated the 9,679 data diffs rather than reading each file. The transform is verified clean and self-consistent: 120,054 legacy `MediaIngredientMech:NNNNNN` ids removed, 120,170 CHEBI ids added (the +116 delta = new `term` insertions from the name-match gap path), **zero** non-legacy `term` `id:` lines removed (no existing groundings clobbered), ingredient `term`/`preferred_term` labels untouched, 1,506 unmappable files correctly retain the legacy field, and the divergence TSV is well-formed (500 rows, 4 cols). Schema field `MediaIngredientMechChebiTerm` is present in the base ref, so the #34 dependency is satisfied. The grounding-audit + backlog reports (G21–G23) document pre-existing upstream mis-groundings, not regressions introduced here.

(GitHub caps the PR diff at 3,000 files; `scripts/` and `reports/` sort after the 9,600+ `data/normalized_yaml/` files, so inline anchoring on the script is not possible.)

**RISK (latent, did not manifest this run) — `scripts/migrate_legacy_mim_to_chebi.py:100` (`_rewrite_entry`, the `if k == "term" and set_term is not None: rebuilt[k] = set_term` branch).** This overwrites the ingredient's existing `term` whenever `set_term` is set, but `set_term` is produced in `_resolve_chebi` (lines 73–87) for *any* term whose id does not start with `CHEBI:` — which includes a term already grounded to FOODON / NCIT / `kgmicrobe.compound:`, not only the `term`-absent gap the inline comment claims ("close grounding gap (term was absent here)"). On a future corpus where a legacy-linked ingredient carries a non-CHEBI grounding, a 0.95-fuzzy MIM name match would silently replace that deliberate grounding with a CHEBI. In this PR it was safe — the aggregate diff shows 0 non-legacy `id:` removals — but to make the script reusable, gate the `term` overwrite on `term` actually being absent/empty (e.g. only build `set_term` when `not cid`) and fix the misleading comment.

No blocking issues; the data transform itself is correct.

realmarcin (commented):
## Review: legacy MediaIngredientMech → CHEBI link refresh

The bulk mechanical rewrite (the ~99.8% "CHEBI present" path) is sound: it is id-safe (uses each ingredient's own CHEBI id + label, does not import MIM's `preferred_term`), in-place / key-order preserving, idempotent (migrated files no longer contain `mediaingredientmech_term:` so re-runs are no-ops), and byte-stable. I verified across the full diff that all 120,054 added `mediaingredientmech_chebi_term` ids are `CHEBI:*` (0 non-CHEBI), no legacy fields are re-added, and the new schema class `MediaIngredientMechChebiTerm` exists (from #34). **However, the ~116 "CHEBI-gap" name-match fills use fuzzy matching at threshold 0.95 and write an authoritative `term` grounding from those hits, mis-grounding at least three chemically distinct ingredients.** Inline comments are on the script; the data-file evidence is in the body (the affected ingredient lines are spread across thousands of files and are not individually anchorable).

### BUG
- `scripts/migrate_legacy_mim_to_chebi.py:75,88` — gap-fill uses fuzzy match (default 0.95) to write a new `term` grounding; mis-grounds near-identical formula strings (see inline).

### RISK
- Gap-fill systematically coalesces DL → L stereoisomers (`DL-Tyrosine` → CHEBI:17895 L-Tyrosine ×15, `DL-threonine` → CHEBI:16857 L-Threonine). Distinct stereochemistry; confirm this is acceptable mapping policy.
- For gap-fill cases where the ingredient already had a non-CHEBI `term` (FOODON/ENVO/NCIT), the fuzzy match still adds a `mediaingredientmech_chebi_term` link to the matched MIM entity — a wrong fuzzy hit yields a wrong cross-reference (does not corrupt grounding, but is asserted as a curated correspondence).

### NIT
- The matched `match_method` (fuzzy_x.xx vs exact_name vs synonym) is computed as `_mode` but discarded — no audit trail records which of the ~116 gap fills were fuzzy, making the risky ones hard to find post-hoc. Record it in the `curation_history` note or the divergence report.
- Schema `MediaIngredientMechChebiTerm` is documented as carrying "the MIM `preferred_term` as label", but the migration deliberately writes the CultureMech-side label instead. Reconcile the schema doc with the actual (intended) behavior.

**BUG — confirmed chemical mis-groundings written by the gap-fill fuzzy match (verified against local CHEBI sqlite via OAK):**
- `Na2SiO3 x 5 H2O` (sodium silicate) → `term: CHEBI:86477` which is **sodium sulfite** — files: `data/normalized_yaml/bacterial/DSMZ_1828_MWC.yaml`, `data/normalized_yaml/bacterial/mwc.yaml`.
- `Na2SeO4 x 10 H2O` (sodium selenate) → `term: CHEBI:32586` which is **sodium sulfate decahydrate** — files: `halorussus_varius_medium_msc_15_2.yaml`, `verrucomicrobia_medium.yaml`.
- `Pyridoxine dihydrochloride` → `term: CHEBI:131532` which is **pyridoxamine dihydrochloride** (different B6 vitamer) — files: `medium_for_strain_c_g2.yaml`, `medium_for_strain_c_l2.yaml`.

Each is a brand-new `term:` grounding that did not exist before this PR (regression), and the corresponding `mediaingredientmech_chebi_term` link is wrong too. Fix at source: restrict gap-fill `term` minting to exact-name/synonym matches (`fuzzy_threshold=1.0`), or gate fuzzy hits behind an element/first-token equality check; then re-run and correct the ~6 emitted files above.

**Note (benign):** most of the other ~100 fuzzy gap fills are harmless typo/i18n normalizations (`Vitamine B12`→Vitamin B12, `L--Asparagine`→L-Asparagine, double-dash artifacts). The transform invariants otherwise check out; aside from the fuzzy-fill bug this is clean.

**BUG — root cause in `scripts/migrate_legacy_mim_to_chebi.py` (could not anchor inline; file is beyond GitHub's 3000-file diff cap):**
- `:75` `match = loader.find_match(ing.get('preferred_term',''), None)` — runs with default `fuzzy_threshold=0.95`, so the gap-fill is NOT exact/synonym-only as described; one-element-symbol formula differences cross 0.95. Fix: pass `fuzzy_threshold=1.0` (exact/synonym only) or gate fuzzy hits behind a first-token/element equality check.
- `:88` `set_term = {'id': str(mc), 'label': str(term_label)} if not cid else None` (written at `:108`) — mints the ingredient's authoritative `term` from that possibly-fuzzy match, so a wrong hit creates a brand-new incorrect grounding (regression). Fix: only set `term` from exact/synonym matches; for fuzzy hits add at most the link with a review flag and leave `term` unset.

realmarcin (commented):
## Summary

Large, mechanical, **id-safe** migration (9,671 files / ~120k links). I reviewed the script + the loader it calls, sampled diffs, and ran corpus-wide and OAK-sqlite spot checks rather than reading all files. The transform holds up well: it reuses each ingredient's **own** CHEBI `term` (no grounding overwrites), the fuzzy gap-fill is genuinely disabled, the new field carries **0** non-CHEBI ids across all 120,054 occurrences, and the 1,506 unmappable files correctly retain the legacy field. Spot-checked gap-fills are correct (`Na2SiO3 x 5 H2O` -> CHEBI:60720 sodium silicate; `Vitamine B12` -> CHEBI:176843). Pre-existing CHEBI mis-groundings (pyridoxine / CHEBI:131531 = pyridoxamine, CHEBI:37583 = trisodium phosphate, CHEBI:32149 = sodium sulfate) are honestly flagged for separate audit and are **not** introduced by this PR.

_(Inline anchors were rejected by the API at HTTP 422 because the PR diff is too large to resolve line positions, so the two script findings are inlined here with file:line.)_

### Verified invariants
- **No grounding overwrite**: `mediaingredientmech_chebi_term` always uses the ingredient's own `term.id` (CHEBI-present branch); existing non-CHEBI groundings (FOODON/ENVO/NCIT) are preserved (`set_term` only fires when `cid` is empty).
- **Fuzzy double-guarded**: `find_match(..., fuzzy_threshold=1.01)` can never return a fuzzy hit (max score 1.0), AND the result is filtered to `match_method in (exact_name, synonym)`. The hardened gap-fill is intact at head.
- **Surgical diffs**: sampled files round-trip byte-identically; only the link-field rename, optional gap-fill `term`, and one `curation_history` entry change.
- **Idempotent**: re-runs skip files lacking the exact legacy key; the substring file filter does not false-match `mediaingredientmech_chebi_term`.
- **No double-key clobber**: base branch had 0 pre-existing `mediaingredientmech_chebi_term`, so the new field cannot collide with an existing one.

### Findings

**RISK - scripts/migrate_legacy_mim_to_chebi.py:99** - the gap-fill path accepts `synonym` matches and then *mints a new authoritative* `term` grounding (`set_term`) on the ingredient. Exact-name hits are safe, but synonym hits are only as reliable as MIM's synonym lists - a noisy MIM synonym can ground a chemically distinct compound, and `term` drives downstream KG chemistry/role inference. The stereo guard below only catches DL/D/L prefixes, not vitamer/isomer confusions (e.g. pyridoxine vs pyridoxamine, the exact class flagged in chebi_grounding_audit.md). Scope is bounded (~116 newly-minted `term:` blocks corpus-wide), so this is low-volume but high-per-record impact. Concrete fix: before merge, dump the minted-term set (ingredient name, match_method, minted CHEBI id+label) to a TSV and spot-review the `synonym`-method rows specifically; consider gating synonym minting behind a manual-review flag while leaving exact-name auto-applied.

**NIT - scripts/migrate_legacy_mim_to_chebi.py:78** - `cid = str(term.get("id")) if term else ""` yields the literal string `"None"` when a `term` dict exists but lacks an `id` key. That makes `cid` truthy, so `not cid` at line 117 is False and the gap-fill is skipped - a malformed/partial term (present but ungrounded) is silently treated as already-grounded and never repaired. It errs safe (no clobber), but a fillable gap is missed. Fix: `cid = str(term.get("id") or "")` so a missing/empty id normalizes to `""`.

Both findings above (moved here from inline due to the 422). All corpus-level checks passed: 0 non-CHEBI ids in the new field (120,054 occurrences), 1,506 files retain the legacy field (matches the ~1,509 unmappable claim), and the `mim_grounding_divergences.tsv` / `chebi_grounding_audit.md` correctly characterize the flagged mis-groundings as pre-existing rather than migration-introduced. No correctness regression found in the bulk transform.

PR #41, opened 2026-06-09
Add id-label-correspondence skill + id↔label QC gate

realmarcin:
Installs the shared **id↔label correspondence** checker (vendored byte-identical across the Mech repos) and wires it as a QC gate.

- `scripts/validate_id_label_correspondence.py` — verifies `(id, label)` pairs in data products correspond to the ontology via OAK.
- **schema**: `Term.label` bound to `rdfs:label` as the **canonical** ontology label for `term.id` (the sibling `preferred_term` holds the free/source name); adds the `rdfs` prefix.
- **`project.justfile`**: `validate-terms-all` (recipe `term.label` gate via `linkml-term-validator --labels`), `validate-products`, `report-label-drift`.
- **CI**: `.github/workflows/label-correspondence.yaml`.
- conf/docs/skill files; `.gitignore` excludes the regenerable `cache/`.

Independent of the grounding PR stack (off `main`), but directly complementary — it validates that CHEBI ids carry their correct labels, which is exactly the **G25 kg_fallback** failure mode (e.g. "Distilled water" tagged `CHEBI:6636` *magnesium dichloride*).


scripts/validate_id_label_correspondence.py
```diff
@@ -0,0 +1,341 @@
...
+    reader = csv.DictReader(lines, delimiter=delim)
+    return list(reader.fieldnames or []), list(reader)
+
+
+def iter_tabular(path: Path, fmt: str, pairs: list[list[str]]) -> Iterator[tuple[str, str, str]]:
+    """Yield (locator, id, label) for each configured id/label column pair."""
+    fields, rows = _read_tabular_rows(path, fmt)
+    field_set = set(fields)
+    usable = [(i, l) for (i, l) in pairs if i in field_set and l in field_set]
+    for n, row in enumerate(rows, start=2):  # row 1 is the header
```
realmarcin: NIT: the tabular row locator is computed against the comment-stripped line list (`_read_tabular_rows` at line 179 drops every `#` line), so `enumerate(rows, start=2)` reports "row 2" for the first data row even though the SSSOM files carry a ~17-line `#` prelude — the locator is off by the prelude length, pointing a curator ~17 lines too early when triaging `reports/label_drift.tsv`. Fix: capture original 1-based line indices while filtering (e.g. keep `(lineno, line)` pairs, or have `iter_tabular` reference the true file line) so the locator matches the actual file.

conf/id_label_targets.yaml
```diff
@@ -0,0 +1,33 @@
...
+    glob: "data/normalized_yaml/**/*.yaml"
+    policy: canonical
+    pairs:
+      - [id, label]
+
+  # data products: CHEBI mapping SSSOM exports — canonical OR synonym
+  - name: sssom
+    kind: tabular
+    format: tsv
+    glob: "output/*.sssom.tsv"
```
realmarcin: NIT: `output/*.sssom.tsv` matches all 18 SSSOM files currently in `output/`, including intermediate/legacy exports (`_clean`, `_exact`, `_enriched`, `_re_enriched`, `_unicode_fixed`, `_with_gases`, …), not just the canonical product. Engine B will re-check the same (id,label) pairs across every derived file, inflating the drift report with duplicate noise. Consider narrowing to the canonical artifact (e.g. `_final` / `_clean`) or a single named product.

project.justfile
```diff
@@ -785,6 +785,32 @@ validate-references file:
...
 
+# id↔label gate (Engine A): the schema now binds every descriptor term slot,
+# so --labels verifies term.label is the CANONICAL ontology label for term.id
+# across all recipe files. Fails (non-zero) on any label drift.
+[group('QC')]
+validate-terms-all:
+    #!/usr/bin/env bash
+    set -uo pipefail
+    rc=0
+    for file in data/normalized_yaml/*/*.yaml; do
```
realmarcin: NIT: Engine A globs `data/normalized_yaml/*/*.yaml` (exactly one dir level) while the Engine B config (`conf/id_label_targets.yaml`) uses `data/normalized_yaml/**/*.yaml` (recursive). They cover the same files today (no nesting deeper than `<category>/`), but if a deeper subdir is ever added Engine A will silently skip it while Engine B checks it — the two gates would disagree. Use `**/*.yaml` (or `find ... -name '*.yaml'`) here for parity.

realmarcin (commented):
Solid, well-scoped addition. The id↔label gate is wired correctly: `schema_path`/`oak_config` justfile vars resolve, the `validate-terms-all` invocation matches the existing `validate-terms` recipe exactly, the schema `rdfs` prefix + `slot_uri: rdfs:label` + range-less bindings are consistent, and the CI workflow is genuinely non-blocking (report-only Phase 1) with OAK sqlite caching. The classifier logic (verdict ordering, synonym scope, allowlist-gated adapter loading) is sound and the YAML walker robustly matches nested `term.{id,label}` blocks. No correctness bugs found — three minor NITs below for triage ergonomics.

scripts/validate_id_label_correspondence.py
```diff
@@ -0,0 +1,461 @@
...
+
+
+def iter_tabular(path: Path, fmt: str, pairs: list[list[str]]) -> Iterator[tuple[str, str, str]]:
+    """Yield (locator, id, label) for each configured id/label column pair."""
+    fields, rows, line_numbers = _read_tabular_rows(path, fmt)
+    field_set = set(fields)
+    # Keep any pair whose id column exists; a missing label column is allowed
+    # so an id present without its label still gets an EMPTY_LABEL verdict
+    # rather than being silently dropped.
+    usable = [(i, l) for (i, l) in pairs if i in field_set]
```
realmarcin, line 368: RISK: When a configured pair's ID column is absent from a product file, the pair is dropped from `usable` and only a stderr warning is emitted — the exit code stays 0. In enforce mode (`just validate-products`) an SSSOM column rename/drop (e.g. `subject_id`) therefore checks nothing and greens silently. Fix: record an ERROR-class verdict (e.g. add a `MISSING_COLUMN` verdict to `_ERROR_VERDICTS`) when `rows` exist but a configured pair can't be applied, so an enforce run exits non-zero instead of passing.

scripts/validate_id_label_correspondence.py
```diff
@@ -0,0 +1,461 @@
...
+        policy = target.get("policy", "canonical_or_synonym")
+        scope = target.get("synonym_scope", default_scope)
+        pairs = target.get("pairs", [])
+        exclude_keys = frozenset(target.get("exclude_keys", []) or [])
+        globs = target.get("glob")
+        glob_list = globs if isinstance(globs, list) else [globs]
+        paths: list[Path] = []
+        for g in glob_list:
+            paths.extend(sorted(REPO_ROOT.glob(g)))
+        if not paths:
```
realmarcin, line 502: RISK: A target whose glob matches zero files prints a note and `continue`s with exit 0. `just validate-products` enforces against `output/culturemech_chebi_mappings.sssom.tsv`; if that product hasn't been built yet (fresh checkout, reordered CI), the gate passes green while validating nothing — a false-green. Fix: add a per-target `required: true` and, in enforce mode, fail (non-zero) when a required target matches no files.

scripts/validate_id_label_correspondence.py
```diff
@@ -0,0 +1,461 @@
...
+    # Strip only the contiguous prelude of ``#`` lines at the top of the file.
+    start = 0
+    while start < len(raw) and raw[start][1].lstrip().startswith("#"):
+        start += 1
+    body = raw[start:]
+    if not body:
+        return [], [], []
+    lines = [text for _, text in body]
+    # body[0] is the header; data rows begin at the next physical line.
+    data_line_numbers = [phys for phys, _ in body[1:]]
```
realmarcin: NIT: `csv.DictReader` skips blank rows, but `data_line_numbers` is built from every line in `body[1:]`, so a blank line within the data region shifts every subsequent locator one (or more) lines early. Only the multiline-quoted-field desync is documented in the docstring. Fix: drop blank lines when building `data_line_numbers` (or derive the locator from the reader's `line_num`) so reported `line N` stays accurate.

project.justfile
```diff
@@ -785,6 +785,34 @@ validate-references file:
...
+    for file in data/normalized_yaml/**/*.yaml; do
+        [ -e "$file" ] || continue
+        uv run linkml-term-validator validate-data "$file" -s {{schema_path}} -t MediaRecipe --labels -c {{oak_config}} || rc=1
+    done
+    exit $rc
+
+# id↔label gate (Engine B): verify (id,label) pairs in DATA PRODUCTS
+# (output/*.sssom.tsv) correspond to the ontology. Exits 2 on any mismatch.
+[group('QC')]
+validate-products:
```
realmarcin, line 814: NIT: This recipe loads the full `conf/id_label_targets.yaml`, which includes the `normalized_yaml` target (policy: canonical) in addition to the SSSOM target — so `validate-products` also enforce-checks recipe YAML `term.label`s, not just `output/*.sssom.tsv` as the comment states. Since the PR docs note many recipe terms still have empty `label: ''`, this will EMPTY_LABEL-fail and double-enforces what Engine A (`validate-terms-all`) already covers. Fix: narrow this recipe to the product target(s), or update the comment to say it gates YAML + products.

realmarcin (commented):
Sanctioned review pass (replacing Copilot). Overall this is a clean, well-documented addition: the vendored `scripts/validate_id_label_correspondence.py`, the `slot_uri: rdfs:label` + range-less `binds_value_of: id` schema bindings, OAK adapter/synonym usage, the `LOAD_FAILED` vs `None` distinction (so a broken-but-configured adapter fails rather than silently skipping), and the report-first rollout all look correct. The findings below are two enforce-path false-green vectors plus two clarity/locator nits; none affect the Phase-1 report-only CI gate.

## RISK
- `scripts/validate_id_label_correspondence.py:267` — a configured pair whose ID column is missing is dropped from `usable` with only a stderr warning; enforce mode still exits 0, so SSSOM column drift greens silently.
- `scripts/validate_id_label_correspondence.py:365` — a target glob matching zero files is skipped with exit 0, so `just validate-products` passes green when the SSSOM product hasn't been built (validates nothing).

## NIT
- `scripts/validate_id_label_correspondence.py:255` — blank lines among data rows desync physical-line locators (DictReader skips empty rows; `data_line_numbers` counts them).
- `project.justfile:807` — `validate-products` loads the full config, so it also enforce-checks recipe YAML labels, not just `output/*.sssom.tsv` as the comment says.

None — all findings anchor to added lines.

src/culturemech/schema/culturemech.yaml
```diff
@@ -684,12 +689,20 @@ classes:
       term:
         description: Primary ontology grounding (CHEBI, FOODON, UBERON, ENVO, or upstream mediadive.compound).
         range: ChemicalEntityTerm
+        # id↔label gate: range-less binding → `linkml-term-validator --labels`
+        # verifies term.label is the canonical OBO label for term.id.
+        bindings:
```
realmarcin: BUG: this `term` binding (and the `chebi_term` one at :703) makes `linkml-term-validator --labels` enforce that `term.label` equals the canonical OBO label. But the data deliberately carries curator labels here (CHEBI:59148→`NaNO3`, CHEBI:131527→`K2HPO4`), and Engine B's `label_waived_keys: [term, chebi_term]` exists to accept exactly those. So Engine A flags as MISMATCH the rows Engine B passes — and since `just validate <file>` (project.justfile:759) already runs `--labels`, this is a live regression, not a Phase-2-only one. Pick one policy: drop the bindings on `term`/`chebi_term`, or relabel and drop the waiver — both gates can't ship contradicting each other.

conf/id_label_targets.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - term
+      - chebi_term
+
+  # data products: CHEBI mapping SSSOM export — canonical OR synonym.
+  # Narrowed to the canonical base artifact only; `output/*.sssom.tsv` also
+  # matched ~18 intermediate/legacy exports (clean/enriched/exact/final/...).
+  - name: sssom
+    kind: tabular
+    format: tsv
+    glob: "output/culturemech_chebi_mappings.sssom.tsv"
```
realmarcin, line 47: RISK: this `sssom` product target has no `required: true`, so the new MISSING_GLOB guard never fires for it. If `output/culturemech_chebi_mappings.sssom.tsv` isn't built, `validate-products` exits 0 (`- sssom: no files match`) — the exact false-green `required` was added to catch. Add `required: true`.

conf/id_label_targets.yaml
```diff
@@ -0,0 +1,51 @@
...
+    # PRODUCT DECISION (item 3c): the `term` and `chebi_term` blocks pair an
+    # OBO id (CHEBI/…) with a CURATOR-INTENDED formula/common-name label —
+    # e.g. CHEBI:15377 "Distilled water", CHEBI:26710 "NaCl",
+    # CHEBI:63036 "KH2PO4". We KEEP those labels (they are the human-facing
+    # recipe names) rather than relabel ~14,900 rows to the OBO canonical form,
+    # so these slots are EXEMPT from canonical-label matching. They are STILL
+    # id-existence checked (a bogus CHEBI id under `term`/`chebi_term` still
+    # surfaces as ID_NOT_FOUND); only the canonical-label match is waived.
+    label_waived_keys:
+      - term
```
realmarcin, line 38: RISK: `label_waived_keys` is matched by bare parent-key name in `_walk_yaml`, so `term` also waives organism `term`, environment `term`, and cofactor `term` blocks — broader than the chemistry-only rationale, and contradicting skill.md (organism/environment `term.label` should be canonical). Meanwhile `precursor_term`/`substrate_terms` are NOT waived, so they get canonical-checked despite the same curator-label style. Key the waiver on slot path or (key, range_class) instead of bare name.

conf/id_label_targets.yaml
```diff
@@ -0,0 +1,51 @@
...
+    kind: yaml
+    glob: "data/normalized_yaml/**/*.yaml"
+    policy: canonical
+    pairs:
+      - [id, label]
+    # mediaingredientmech_chebi_term.label is intentionally MIM's preferred_term
+    # (not the OBO canonical label), so excluded from the canonical gate to
+    # avoid false MISMATCHes.
+    exclude_keys:
+      - mediaingredientmech_chebi_term
```
realmarcin, line 28: NIT: `mediaingredientmech_chebi_term` is not a slot in the CultureMech schema (only `mediaingredientmech_term` exists), so this `exclude_keys` entry never matches — dead config. Use the actual slot name if exclusion was intended.

realmarcin (commented):
Sanctioned review pass. The vendored validator (Engine B) is well-hardened and re-verified at head: the sha256 pin matches the script byte-for-byte (0a446a8…), and the MISSING_COLUMN / ADAPTER_ERROR / LOAD_FAILED / EMPTY_ADAPTER / physical-line-locator paths all read correctly. The substantive issue is a contradiction between the two engines this PR introduces on the `term`/`chebi_term` slots, plus a couple of config-wiring gaps that defeat the very guards the script added.

## BUG
- **src/culturemech/schema/culturemech.yaml:694 / :703** — Engine A and Engine B directly contradict each other on `term`/`chebi_term`. The schema now binds these slots (`binds_value_of: id`, REQUIRED), so `linkml-term-validator --labels` will compare `term.label` against the canonical OBO label. But the live data deliberately carries curator labels there — e.g. `CHEBI:59148`→`NaNO3`, `CHEBI:131527`→`K2HPO4` (canonical = "sodium nitrate" / "potassium hydrogen phosphate") — and Engine B's `label_waived_keys: [term, chebi_term]` exists precisely to accept them. So Engine A flags as MISMATCH the exact rows Engine B passes. This is not just a latent Phase-2 problem: `just validate <file>` (project.justfile:759) and `just validate-terms-all` (:800) already run `--labels`, so this regresses the existing per-file validate workflow on every recipe with a curator label. Resolve the policy in one place: either drop the bindings on `term`/`chebi_term` (keep them on organism/environment/precursor/substrate where labels are canonical), or relabel and abandon the waiver — but the two gates cannot both ship as-is.

## RISK
- **conf/id_label_targets.yaml:47** — the `sssom` product target does not set `required: true`, so the newly added MISSING_GLOB guard never fires for it. If `output/culturemech_chebi_mappings.sssom.tsv` hasn't been built, `just validate-products` still exits 0 (`- sssom: no files match …`) — the exact false-green the `required` knob was added to prevent. Add `required: true` to this target.
- **conf/id_label_targets.yaml:38** — `label_waived_keys` is matched by bare parent-key name (`_walk_yaml`: `k in label_waived_keys`), so `term` waives not only the ingredient `term` but also organism `term`, environment `term`, and cofactor `term` blocks — broader than the chemistry-only "formula/common-name" rationale, and contradicting skill.md which states organism/environment `term.label` must be canonical. A wrong-but-resolvable NCBITaxon/ENVO label in a `term` block passes as OK_ID_ONLY. Conversely `precursor_term`/`substrate_terms` are NOT waived, so they get canonical-checked despite holding the same curator-label style — inconsistent treatment. Consider keying the waiver on (parent_key, range_class) or an explicit slot path rather than bare name.

## NIT
- **conf/id_label_targets.yaml:28** — `exclude_keys: [mediaingredientmech_chebi_term]` references a slot that does not exist in the CultureMech schema (only `mediaingredientmech_term` exists). It's dead config; the comment's intent (exclude MIM's preferred_term label) isn't actually achieved here. (Harmless in practice because `MediaIngredientMech:` has no adapter → SKIPPED_NO_ADAPTER, but the key name is wrong.)

- **project.justfile:759 (`just validate`) — not in this hunk** — this pre-existing recipe runs `linkml-term-validator … --labels`, which is what makes the BUG above a live regression rather than a Phase-2-only concern. Worth a note in the PR body so reviewers know the contradiction bites today.
- **Tabular MISSING_COLUMN only fires when `rows` is non-empty** — a header-only SSSOM (zero data rows) with a renamed id/label column would not surface MISSING_COLUMN. Combined with the missing `required: true`, an empty-but-present export is a blind spot. Low priority.

realmarcin:
## Review findings addressed

**BUG — Engine A/B contradiction on `term`/`chebi_term` (schema:694/703):** resolved by **dropping the canonical-label bindings** on `term` and `chebi_term`. They carry curator formula/common-name labels (NaNO3, K2HPO4, 'Distilled water') that are deliberately non-canonical, which is exactly what Engine B's `label_waived_keys` accepts. Canonical-label bindings stay on organism/environment/precursor/substrate terms. The two engines no longer contradict, and `just validate <file>` no longer regresses on curator-labelled rows.

**RISK fixes**
- `conf:47` / zero-file glob: added `required: true` to the sssom product target so enforce mode fails (MISSING_GLOB) on a missing product instead of greening on nothing.
- `conf:38` waiver breadth: the label waiver is now scoped to **CHEBI** ids, so a bare `term` waiver no longer silently exempts organism (NCBITaxon) / environment (ENVO) term labels — those stay canonical-checked.
- MISSING_COLUMN drift (validator:365): already an ERROR verdict in the hardened script (enforce exits non-zero).

**NIT fixes**
- Blank-line locator desync (validator:340): truly-empty lines are now excluded from tabular line numbers (DictReader skips them), so a blank data row no longer shifts later locators. Verified on a fixture.
- SSSOM prelude locator off-by-~17 (validator:191): already fixed — locators use true physical line numbers.
- `output/*.sssom.tsv` matched 18 files (conf:29): already narrowed to the single canonical product.
- Engine-A glob parity (justfile:796): already `**/*.yaml`.
- `validate-products` comment (justfile:807): corrected to state it gates recipe YAML + SSSOM products.
- `exclude_keys: mediaingredientmech_chebi_term` (conf:28): no longer dead config — that slot was added in #34 and is now live.

Re-pinned `scripts/.validate_id_label_correspondence.sha256` after the validator edit. Note: the vendored validator now differs from sibling Mech repos and should be synced there (PR #47 carries this hardening forward in-repo).

PR #42, opened 2026-06-09
G25 Phase 1: re-ground 15,319 kg_fallback chebi_terms via corpus consensus (0 API)

realmarcin:
First phase of the G25 kg_fallback remediation — the **zero-external-call** tier from the scope.

## Problem
The `chebi_term` field carried a large low-confidence `kg_fallback` layer, frequently wrong — flagship case: **"Distilled water" → `CHEBI:6636` magnesium dichloride, ×3,951**.

## Fix (consensus borrow, no OAK/OLS)
`scripts/migrate_kg_fallback_to_consensus.py`: for each ingredient whose only grounding is a low-confidence `chebi_term` and whose label has a **unanimous** reliable grounding elsewhere in the corpus (primary CHEBI `term` or `exact_match` chebi_term), replace it with that consensus id+label and `match_type: corpus_consensus`. Labels with **conflicting** reliable ids (the G23 same-name-multiple-CHEBI cases, e.g. hydrate variants) are **deferred to Phase 2** (OAK + OLS).

## Result
- **15,319 chebi_terms re-grounded across 4,497 records** (1,361 unanimous labels).
- Low-confidence layer **23,998 → 8,679**; `kg_fallback` **21,146 → 7,418**.
- **"Distilled water" on CHEBI:6636: 3,951 → 0.**
- Verified: water→`CHEBI:15377`, CaCl₂·2H₂O→`CHEBI:86158` (was Vitamin B12), agar/KH₂PO₄/Na₂S correct; yeast extract correctly deferred.
- Corpus validates **0 ERROR rows** (closed mode).

## Schema
Adds `corpus_consensus` to `TermMatchTypeEnum` (+ regenerated dataclasses); the G23 probe now counts it as reliable. G23 Signal A held at 102 — no new conflicts (consensus agrees by construction), so the CI gate baseline is unchanged.

## Next
Phase 2 (OAK `sqlite:obo:chebi` + OLS4, agreement-required) for the ~remaining labels; Phase 3 de-ground the unresolved. See `reports/g25_kg_fallback_remediation_scope.md`.

## Dependency
⚠️ Stacked on **#40** → #39 → #38 → #34.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #40, opened 2026-06-09
Grounding follow-ups: traversal blind spot + G22/G24 re-grounding + G23 CI probe

realmarcin:
Resolves the follow-ups from the grounding work, in order.

## 1. Traversal blind spot (issue 1)
Standalone solution records keep ingredients in a top-level `composition:` list. Audited all bulk mutators; fixed the two that missed it:
- **`mediaingredientmech_linker.py`** (production MIM enricher) — now walks `ingredients` + top-level `composition` + `solutions[].composition`.
- **`remove_invalid_chebi_ids.py`** — same.
(`enrich_solutions_with_chebi.py` / `cleanup_residual_errors.py` already handled it; `enrich_with_chebi.py` is covered by the solutions enricher.) The MIM-migration half landed on #38.

## 2. G22 — phosphate speciation
Split `CHEBI:37583` (trisodium phosphate): monobasic → **CHEBI:37585** (156), dibasic → **CHEBI:34683** (33). No genuine trisodium. 111 records.

## 3. G24 — shared-id garbage (targeted)
Re-grounded the *audited* garbage on `CHEBI:32149` (lactate/propionate/NiCl₂/malate, verified targets) and de-grounded complex media on `CHEBI:15978` (51 refs).
**Rejected the blanket exact-label-to-MIM remap as unsafe** — it would touch 6,519 occurrences, but MIM itself mis-grounds glycerol and casamino acids, so it would *undo G21* and propagate MIM's errors. The rest needs a trusted CHEBI source + curation.

## 4. G23 — grounding-consistency QC probe + CI gate
`scripts/audit_chebi_consistency.py` + `just check-chebi-grounding` + `.github/workflows/chebi-consistency.yaml`. Self-consistency check (no external truth): flags names grounded to >1 CHEBI on the **reliable** layer; gates at baseline (102) so new inconsistencies fail.

## New finding → G25
The probe surfaced a large low-confidence layer: **21,146 `kg_fallback` + 2,746 ambiguous `chebi_term` fields (61% of all chebi_terms)**, frequently wrong (e.g. "Distilled water" → CHEBI:6636 *magnesium dichloride*, confidence 0.7). Primary `term` grounding is unaffected; flagged for a dedicated pass.

## Dependency
⚠️ Stacked on **#39** → #38 → #34. Bases retarget up the stack as each merges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #43, opened 2026-06-09
G25 Phase 2: resolve kg_fallback chebi_terms via OAK + OLS

realmarcin:
Second phase of the G25 kg_fallback remediation — external grounding via **OAK `sqlite:obo:chebi` + OLS4**, with a corpus-majority disambiguator and mixture de-grounding.

## Outcomes (3,607 chebi_terms across 2,635 records)
| outcome | n | how |
|---|---|---|
| corpus-majority | 1,612 | ≥90% corpus reliable majority, cross-checked against OAK∪OLS exact (hydrate-safe) |
| ontology-exact | 224 | OAK single exact ∈ OLS exact set (clean chemicals: KNO₃→63043, Na₂SeO₄→77775) |
| de-grounded | 1,771 | complex mixtures with no single CHEBI (peptone/extract/broth/sea water/serum…) |
| **flagged** | 281 labels | left as-is, written to `kg_fallback_phase2_curator_review.tsv` — **no guessing** |

## Hydrate bug caught before applying
Name-normalisation drops the water-of-crystallisation (`MgSO₄·7H₂O` → "magnesium sulfate"), so an ontology lookup would ground **3,077 heptahydrate entries to the hexahydrate**. Fix: **corpus-majority-first** (ontology cross-checked) keeps the correct **CHEBI:31795**; unconfirmed hydrates are flagged, not guessed.

## Net effect (Phase 1 + 2)
- `kg_fallback` chebi_terms **21,146 → 3,818 (82% reduction)**; low-confidence layer **23,998 → 5,072**.
- Catastrophic errors eliminated ("Distilled water"→MgCl₂ ×3,951, CaCl₂→Vitamin B12).
- Corpus validates **0 ERROR**; G23 Signal A unchanged (102).

## Remaining (Phase 3)
The 281 flagged labels (mostly stereochem-prefixed organics whose existing grounding is plausibly correct but unconfirmed) — needs curator judgement, not automation.

## Dependency
⚠️ Stacked on **#42** → #40 → … → #34.


scripts/migrate_kg_fallback_phase2_oak_ols.py
```diff
@@ -0,0 +1,276 @@
...
+        if corpus_rel:
+            total = sum(corpus_rel.values())
+            c, n = corpus_rel.most_common(1)[0]
+            if total >= 5 and n / total >= 0.90 and (c in oak or c in ols):
+                return "resolve", c, label, "corpus_consensus"
+
+        # Path A — single precise OAK hit confirmed by OLS (non-hydrate only).
+        if not has_hydrate and len(oak) == 1:
+            c = next(iter(oak))
+            if c in ols or not ols:
```
realmarcin: RISK: Path A accepts the single OAK exact hit when `ols` is empty (`c in ols or not ols`). But `ols_exact` swallows network/timeout errors and returns an empty set (the bare `except Exception: pass` around the requests.get above), so a transient OLS4 outage is indistinguishable from a genuine 'no OLS exact' result. During such an outage every non-hydrate single-OAK hit would resolve to `match_type: exact_match` without the advertised OLS cross-check. Consider distinguishing 'OLS errored' from 'OLS returned nothing' (e.g. a sentinel / raise) before treating an empty OLS set as a pass.

scripts/migrate_kg_fallback_phase2_oak_ols.py
```diff
@@ -0,0 +1,276 @@
...
+            if not isinstance(ct, dict) or ct.get("match_type") not in LOW_CONF:
+                continue
+            lk = name_key(ing.get("preferred_term", ""))
+            if lk not in decided:
+                decided[lk] = resolver.resolve(ing.get("preferred_term", ""),
+                                               corpus_rel.get(lk))
+            action, cid, lbl, method = decided[lk]
+            was = str(ct.get("id"))
+            if action == "resolve":
+                ing["chebi_term"] = {"id": cid, "label": lbl,
```
realmarcin: RISK: the resolved `chebi_term.label` is set to `lbl`, which is the ingredient `preferred_term` (e.g. 'Na2SeO4', 'MgSO4 x 7 H2O'), not the canonical CHEBI label. For the `oak_ols_exact` path this writes `match_type: exact_match` with a non-ontology label, diverging from the existing upstream exact_match convention (e.g. CHEBI:131527 → 'dipotassium hydrogen phosphate', CHEBI:32149 → 'sodium sulfate') and from the schema slot `label: Ontology term label`. These rows will fail id↔label correspondence validation. Since OAK already confirmed the id, fetch the canonical label: `lbl = resolver.oak.label(cid) or label` (have `resolve` return the ontology label instead of the input). NIT: `confidence` is also dropped here, though upstream exact_match entries carry `confidence: 1.0`. (The corpus_consensus path inherits the ingredient-name label from Phase 1 precedent, so that's not a new regression — but the same id↔label caveat applies.)

realmarcin (commented):
Reviewed the G25 Phase 2 migration: the new `migrate_kg_fallback_phase2_oak_ols.py` plus ~2,635 YAML regroundings and report TSVs. The overall approach is sound and safety-conscious — corpus-majority-first to dodge the hydrate-normalisation trap, ontology cross-check, mixture de-grounding, and no-guess flagging are all good, and the regenerated `chebi_consistency.tsv` shows the expected consensus consolidation with no new variants or ERROR regressions. Two RISKs and a NIT in the resolver, noted inline.

realmarcin (commented):
## Review — G25 Phase 2 (OAK + OLS kg_fallback resolution)

The migration script is well-structured and the headline hydrate hazard is genuinely handled correctly: corpus-majority-first (Path B) plus the `has_hydrate` guard keeps `MgSO₄·7H₂O` on CHEBI:31795 (verified in `mediadive_1004_Main_sol_493.yaml`). Resolved labels are canonical (written from `oak.label(cid)`, not the ingredient name), prior `confidence` is preserved via the `{**ct, ...}` merge, and the emitted `match_type` values (`exact_match`, `corpus_consensus`) are valid `TermMatchTypeEnum` members. Idempotency holds since resolved/removed terms no longer satisfy the `LOW_CONF` re-entry condition. Two consistency findings below, plus one schema-doc drift not anchorable to the diff.

### RISK
- `scripts/migrate_kg_fallback_phase2_oak_ols.py:194` — the de-ground guard conflates an OLS *error* with a genuine OLS *empty* result, bypassing the exact safeguard the docstring of `ols_exact` promises and that Path A honors at line 190.

### NIT
- `scripts/migrate_kg_fallback_phase2_oak_ols.py:190` — the `(not ols and not ols_errored)` branch resolves on a single OAK hit when OLS returned no exact record, yet still stamps `match_type="exact_match"`, overstating the OAK∩OLS agreement the docstring describes.

- NIT — `src/culturemech/schema/culturemech.yaml:2550` (not in this PR's diff): the `corpus_consensus` enum description still says "unanimous reliable grounding ... without an external lookup" (Phase 1 semantics). Phase 2 reuses this value for a ≥90% strong majority that IS externally cross-checked against OAK∪OLS. Update the description so the enum matches its Phase 2 meaning.

scripts/migrate_kg_fallback_phase2_oak_ols.py
```diff
@@ -0,0 +1,292 @@
...
+        if not has_hydrate and len(oak) == 1:
+            c = next(iter(oak))
+            # Take the "OLS empty" shortcut only when OLS genuinely returned
+            # nothing — never when the OLS call errored/timed out (which must
+            # not silently bypass the cross-check).
+            if c in ols or (not ols and not ols_errored):
+                return "resolve", c, label, "oak_ols_exact"
+
+        # De-ground clear mixtures with no ontology grounding
+        if not oak and not ols and MIXTURE.search(label):
```
realmarcin: RISK: when OLS errored, `ols` is the empty set (None coalesced at line 168) while `ols_errored` is True. This guard tests only `not ols`, so during an OLS outage a mixture-pattern label with no OAK hit is de-grounded (chebi_term removed) without the ontology cross-check this path is meant to perform — exactly the error-vs-empty conflation `ols_exact`'s docstring and Path A (line 190) take care to avoid. Blast radius is small (only the low-confidence chebi_term is removed; the primary `term` is untouched, and true mixtures rarely have an OLS-only exact hit), but it is inconsistent with the stated contract. Fix: add `and not ols_errored` to the guard.

scripts/migrate_kg_fallback_phase2_oak_ols.py
```diff
@@ -0,0 +1,292 @@
...
+            if total >= 5 and n / total >= 0.90 and (c in oak or c in ols):
+                return "resolve", c, label, "corpus_consensus"
+
+        # Path A — single precise OAK hit confirmed by OLS (non-hydrate only).
+        if not has_hydrate and len(oak) == 1:
+            c = next(iter(oak))
+            # Take the "OLS empty" shortcut only when OLS genuinely returned
+            # nothing — never when the OLS call errored/timed out (which must
+            # not silently bypass the cross-check).
+            if c in ols or (not ols and not ols_errored):
```
realmarcin, line 197: NIT: the `(not ols and not ols_errored)` branch resolves on a single OAK exact hit when OLS genuinely returned no exact record at all, then stamps `match_type="exact_match"`. The Path A docstring says resolution requires the id be "in the OLS4 exact set", so this overstates cross-check confirmation for ids OLS never confirmed. It is deliberate (per the comment) and limited to non-hydrate single-OAK hits, but consider a distinct method/confidence (or a doc note) so `exact_match` is not claimed without OLS agreement.

scripts/migrate_kg_fallback_phase2_oak_ols.py
```diff
@@ -0,0 +1,295 @@
...
+        forms = {label, norm}
+        oak = set().union(*(self.oak_exact(f) for f in forms))
+        ols_results = [self.ols_exact(f) for f in forms]
+        ols_errored = any(r is None for r in ols_results)
+        ols = set().union(*(r or set() for r in ols_results))
+        # Hydrate/formula labels lose their water-of-crystallisation under
+        # normalisation, so an ontology lookup on the normalised name can pick
+        # the WRONG hydrate (e.g. "MgSO4 x 7 H2O" -> magnesium sulfate ->
+        # hexahydrate). For these, trust only the corpus majority, never Path A.
+        has_hydrate = bool(re.search(r"x\s*\d+\s*h2?o|hydrate", label, re.I))
```
realmarcin: RISK: hydrate guard only matches the ` x N H2O` form, not the middot/x-times form (MgSO4-middot-7H2O, MnCl2-middot-4H2O, Na2S-middot-9H2O) which is widespread in the corpus. For those has_hydrate=False, so Path A (L185) can ground the water-stripped normalised name to the wrong hydrate when the corpus lacks a >=90%/>=5 majority. Fix: re.search(r"\d+\s*h2?o|hydrate", label, re.I) to catch water-of-crystallisation regardless of separator.

scripts/migrate_kg_fallback_phase2_oak_ols.py
```diff
@@ -0,0 +1,295 @@
...
+        self.oak = get_adapter("sqlite:obo:chebi")
+        self._ols: dict[str, set] = {}
+        self._oak: dict[str, set] = {}
+
+    def oak_exact(self, q):
+        if q in self._oak:
+            return self._oak[q]
+        out = set()
+        try:
+            for cid in list(self.oak.basic_search(q))[:12]:
```
realmarcin: NIT: list(self.oak.basic_search(q))[:12] truncates before the exact-synonym filter, so an exact match ranked beyond the first 12 results is silently missed. Low impact but the cap is a correctness-affecting heuristic.

scripts/migrate_kg_fallback_phase2_oak_ols.py
```diff
@@ -0,0 +1,295 @@
...
+
+import requests
+import yaml
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+NORMALIZED_DIR = REPO_ROOT / "data" / "normalized_yaml"
+CURATOR = "kg-fallback-oak-ols-v1.0"
+LOW_CONF = {"kg_fallback", "synonym_match_ambiguous", "(none)", None}
+RELIABLE = {"exact_match", "corpus_consensus"}
+MIXTURE = re.compile(
```
realmarcin, line 57: NIT: MIXTURE requires a qualifier word, so bare complex materials keep their wrong groundings in the curator-review TSV (activated carbon -> CHEBI:86158 calcium chloride dihydrate, bakers yeast -> CHEBI:86158, beef -> CHEBI:2509 agar). Conservative flagging is intentional; consider adding bare yeast|beef|carbon|soil|meal|brain|albumin to the deground set in Phase 3.

realmarcin (commented):
## G25 Phase 2 review — OAK + OLS kg_fallback grounder

Solid, conservative remediation. I verified the load-bearing safety logic and a sample of the data against the local OAK `chebi.db`, and it holds up: the OLS **error-vs-empty** guards are correctly implemented (both Path A at L185-191 and the de-ground at L199 require `not ols_errored`, so a transient OLS outage cannot be read as "no match"); `confidence` is preserved via the `{**ct, ...}` merge on resolve; labels are canonicalised from OAK (`CHEBI:8806`→`Resazurin` fixed, and **0** `label: CHEBI:*` rows remain in `data/normalized_yaml` at head); and sampled resolved ids verify as real catastrophic-error fixes — `mgso4 x 7 h2o` CHEBI:31206 (ammonium chloride) → CHEBI:31795 (magnesium sulfate heptahydrate), `na-thioglycolate` CHEBI:86158 (calcium chloride dihydrate) → CHEBI:86481 (sodium thioglycolate). One RISK on the hydrate safety-net, plus minor NITs.

### RISK
- **scripts/migrate_kg_fallback_phase2_oak_ols.py:173** — the hydrate guard `re.search(r"x\s*\d+\s*h2?o|hydrate", label, re.I)` only catches the ` x N H2O` textual form, but the corpus also widely uses the middot/`×` form (`MgSO4·7H2O`, `MnCl2·4H2O`, `Na2S·9H2O`, `Na2SeO3·5H2O`, …). For those `has_hydrate=False`, so Path A can fire on the water-stripped normalised name and ground to the wrong (anhydrous/other) hydrate when the corpus lacks a ≥90%/≥5 majority. Corpus-consensus catches the common cases empirically, but the defensive guard is incomplete. Fix: detect water-of-crystallisation regardless of separator, e.g. `re.search(r"\d+\s*h2?o|hydrate", label, re.I)`.

### NIT
- **scripts/migrate_kg_fallback_phase2_oak_ols.py:127** — `list(self.oak.basic_search(q))[:12]` truncates before the exact-synonym filter; an exact match ranked beyond the first 12 results is silently missed. Low impact (exact hits usually rank high), but the cap is a correctness-affecting heuristic worth a comment or a bump.
- **scripts/migrate_kg_fallback_phase2_oak_ols.py:57** — the MIXTURE de-ground pattern requires a qualifier word (`*extract`/`broth`/`serum`/…). Bare complex materials still slip through to the curator-review TSV with their old wrong groundings intact, e.g. `activated carbon`→CHEBI:86158 (calcium chloride dihydrate), `baker's yeast`→CHEBI:86158, `beef`→CHEBI:2509 (agar). Conservative flagging is intentional ("no guessing"), so this is fine to defer to Phase 3, but consider adding bare `yeast|beef|carbon|soil|meal|brain|albumin` to the deground set.

- (NIT, L218) `ct.get("match_type") not in LOW_CONF` — `LOW_CONF` includes `None`, so a `chebi_term` with a *missing* `match_type` is treated as low-confidence and eligible for overwrite/removal. Probably intended, but worth confirming no well-grounded terms lack the field.
- (Observation) De-ground vs flag is OLS-availability dependent: a mixture whose OLS call errors is correctly flagged (kept) rather than de-grounded, so re-runs against a healthy OLS may de-ground more. Safe and idempotent, but results are not fully deterministic across runs. This explains why some `serum`/mixture rows landed in the curator-review TSV instead of being removed.
- (Observation) `build_corpus_reliable` counts prior-run `corpus_consensus` terms into the majority, so repeated runs can reinforce a majority. Not a problem given resolved terms aren't reprocessed (not in `LOW_CONF`), just noting the feedback path.
- Data integrity: spot-checked to_chebi ids all verify in OAK; `mixture_no_chebi` de-grounds (Malt extract→CHEBI:2509 agar removed) are correct; flagged-row current_chebi values are frequently wrong but intentionally left as-is per the stated "no guessing" policy.

realmarcin:
## Review findings addressed

- **OLS error vs empty on Path A** (line ~178): fixed — `ols_exact` returns `None` on error; Path A takes the OLS-empty shortcut only when `not ols and not ols_errored`.
- **De-ground on OLS error** (line ~194): fixed — mixture de-ground guarded by `and not ols_errored`.
- **Non-canonical resolved label** (line ~212): fixed — resolved `chebi_term.label` uses the canonical OAK label (`resolver.oak.label(cid) or lbl`), and the term dict is merged (`{**ct, …}`) so prior `confidence` survives.
- **Hydrate guard missed middot form** (line ~173): fixed — `\d+\s*h2?o|hydrate` now catches `MgSO4·7H2O`/`MnCl2·4H2O`/`Na2S·9H2O`, so Path A no longer risks the wrong hydrate.
- **OAK search truncation** (line ~127): fixed — cap raised 12 → 50 so an exact hit ranked beyond the cap isn't dropped before the exact filter.
- **exact_match claimed without OLS** (line ~190): the Path A comment documents that the OLS-empty shortcut is taken only when OLS genuinely returned nothing (non-hydrate single-OAK hit); behavior is intentional and now explicit.
- **Bare complex materials kept wrong groundings** (line ~57): MIXTURE conservatism is intentional; extending the de-ground set to bare `yeast/beef/carbon/soil/…` is carried into Phase 3 (PR #45).

Consistency stays within the re-baselined 111 (phase-2 OAK/OLS resolution reduced the low-confidence kg_fallback layer from ~8.7k to ~5.1k fields).

PR #45, opened 2026-06-09
G25 Phase 3: LLM-curate 281 flagged labels + MgSO4·7H2O primary-term fix

realmarcin:
Final G25 phase — LLM curation of the labels neither corpus-consensus (P1) nor OAK+OLS (P2) could safely resolve, plus the requested MgSO₄·7H₂O primary-`term` fix.

## Phase 3 — LLM curation (4,972 chebi_terms)
An LLM curator classified the 281 flagged labels in `g25_phase3_llm_curation.tsv`. **Safety pattern: the curator asserts the compound *name*, never a CHEBI id** — the id is resolved by **OAK `sqlite:obo:chebi`** (exact, OLS-cross-checked), so a curator slip can't inject a wrong id.

| outcome | n | how |
|---|---|---|
| resolved | 1,477 | OAK+OLS exact on the curated name, or (for formula-style CHEBI hydrate labels a name-lookup can't match) the corpus's own ≥90% reliable majority |
| de-grounded | 2,276 | mixtures/biologicals/media (yeast extract, peptone, soil, sera, milk, broths…) |
| confirmed | 1,219 | grounding already correct → upgraded to `manual_curated` |

The triage was the key insight: the flagged labels' kg_fallback ids were **wildly wrong** (`(-)-chloramphenicol → alginic acid`, `b12 → nickel dichloride`, `bovine serum albumin → magnesium dichloride`), not "probably correct."

## MgSO₄·7H₂O primary-`term` fix (1,069)
Reliable-layer analogue of G22/G24 (which only touched chebi_term). **1,069** explicitly-heptahydrate ingredients were grounded on the **primary `term`** to `CHEBI:32599` (generic magnesium sulfate) → corrected to **`CHEBI:31795`** (heptahydrate). Generic "Magnesium sulfate" and 6-hydrate labels untouched. (Note: 32599 is the *generic* class, not hexahydrate as first thought.)

## Net G25 (Phases 1–3)
- **Low-confidence chebi_term layer: 23,998 → 100 (99.6% eliminated)**; kg_fallback 21,146 → 100.
- Corpus validates **0 ERROR**; G23 Signal A **102 → 101**; CI gate ratcheted to 101.
- **Remaining:** ~48 low-occurrence exotic-salt labels left flagged for a curator (no ontology name-match, no corpus majority) — genuinely needs human judgement.

## Dependency
⚠️ Stacked on **#43** → #42 → #40 → … → #34.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #46, opened 2026-06-10
Fix primary-term groundings: 483 wrong-id regrounds + 20,857 label canonicalizations

realmarcin:
Stacked on #45 (`g25-phase3-curation`). All of G25 fixed only `chebi_term`; this
PR fixes the **primary `term` field**, which `report-label-drift` surfaced as
still carrying both wrong-ids and non-canonical labels. Two separate, ordered
passes (wrong-id first so canonicalization runs on corrected ids), each its own
commit, corpus validates **0 ERROR** after each.

`scripts/fix_primary_term_grounding.py` classifies every distinct
`(term.id, preferred_term)` pair via OAK `sqlite:obo:chebi`. An action fires
only when the ingredient NAME resolves to a single exact CHEBI — names that don't
resolve are flagged, never guessed.

## (a) Wrong-id reground — `052872c8`
- **483 primary `term` ids re-grounded** across 462 records.
- LLM-curated → OAK-verified, gated by a **13-transition allowlist** of genuine
  cross-compound errors (e.g. Ca-pantothenate off capsaicin, KNO₃ off Renilla
  luciferyl sulfate, citric acid off sodium citrate, Na₂SeO₄ off triisocyanate).
- A `_same_compound` guard prevents the naive-pass trap (16,513 occurrences that
  would have downgraded every correct hydrate id to its generic and reverted the
  MgSO₄·7H₂O fix). Salt/acid downgrades and lateral synonyms excluded.
- Report: `data/import_tracking/reports/primary_term_reground_changes.tsv`.

## (b) Label canonicalization — `efc7239b`
- **20,857 primary-term labels canonicalized** across 6,683 records.
- Fires only where OAK confirms the name resolves to a **single exact CHEBI ==
  the current id** — a correct grounding is never relabeled to a different
  compound, and no unfixed wrong-id is masked. `preferred_term` keeps the source
  string. Top hits: `KH2PO4`→"potassium dihydrogen phosphate", `H3BO3`→"boric
  acid", `NH4Cl`→"ammonium chloride", `K2HPO4`→"dipotassium hydrogen phosphate".
- Report: `data/import_tracking/reports/primary_term_canonicalize_changes.tsv`.

## Left untouched (by design)
- **1,526 flagged primary-term pairs** — formula/recipe strings that don't
  OAK-resolve, or resolve only to a same-compound hydrate variant. Not guessed,
  not masked; candidate pool for a future LLM-curated round.
- `chebi_term`-field label canonicalization is a separate pass.


realmarcin (commented):
## Review: primary-term grounding fixes (verified against local OAK `chebi.db`)

I independently verified this PR against a local OAK CHEBI sqlite (`~/.data/oaklib/chebi.db`) without trusting the asserted ids, and it holds up well. **All 13 reground transitions (483 rows) resolve to the correct compound for the ingredient name**, the canonicalization pass is provably label-only (0 `id:` lines change between the reground and canonicalize commits; 16,990 `label:` lines change), and the two named prior defects are fixed at head. The methodology is conservative and sound — reground is doubly gated (a hardcoded LLM-curated allowlist *and* an OAK single-exact-resolution check + `_same_compound` guard), so it cannot introduce arbitrary new wrong ids; canonicalize only relabels where the name resolves to the *same* current id, so it cannot mask an unfixed wrong id.

### Verification performed
- **483 regrounds / 13 transitions — all CORRECT.** Each from_id/to_id checked against `rdfs_label_statement`: KNO3 CHEBI:58242 (Renilla luciferyl sulfate) → CHEBI:63043 potassium nitrate (309); Na2SeO4 CHEBI:59160 (triisocyanate) → CHEBI:77775 sodium selenate (104); Na2S2O3 → sodium thiosulfate; (NH4)2HPO4 → diammonium hydrogen phosphate; ACES → CHEBI:39061 ACES; Ca-pantothenate CHEBI:3374 (capsaicin) → CHEBI:31345; pyridoxine·2HCl off pyridoxamine·2HCl → CHEBI:189426; citric acid off sodium citrate → CHEBI:30769; K-acetate → CHEBI:32029; Trigonelline·HCl → CHEBI:229203; HEPES off HEIDA → CHEBI:46756. No mis-regroundings found.
- **Named prior defects fixed at head:** MgCl2·6H2O is now CHEBI:86345 *magnesium dichloride hexahydrate* (no MgCl2 ingredient remains on CHEBI:34887 nickel dichloride; the 3,313 remaining CHEBI:34887 occurrences are legitimately bound to `NiCl2 x 6 H2O`). "Sea water" no longer carries CHEBI:75228 sodium lactate (now mediadive.solution:1976; remaining CHEBI:75228 hits are legitimate `Na-DL-lactate`). Note these fixes land via the base branch, not this PR's reground report.
- **Canonicalize ids sampled & correct:** KH2PO4→CHEBI:63036, H3BO3→CHEBI:33118, NH4Cl→CHEBI:31206, KBr→CHEBI:32030, Na2WO4→CHEBI:63940 — all match.

### Findings (all minor; no BUG)
See inline NIT/RISK comments on `scripts/fix_primary_term_grounding.py`.

- **NIT — canonicalize report count vs. actual changes.** `primary_term_canonicalize_changes.tsv` reports 20,857 canonicalizations but only 16,990 `label:` lines actually change in the diff (052872c8..HEAD). The ~3,867 gap are records whose `term.label` already equaled the OBO canonical — the script logs a changelog row on every `action == "canonicalize"` regardless of whether `t["label"]` actually differed, inflating the headline count. Harmless, but the report over-states changes. Fix: only append to `changelog` when `t.get("label") != a`.
- **RISK (out-of-scope / pre-existing) — primary term now diverges from `mediaingredientmech_chebi_term` on cross-compound regrounds.** Example: the `Na2SeO4 x 10 H2O` record's primary `term` is correctly re-grounded to CHEBI:77775 (sodium selenate), but its `mediaingredientmech_chebi_term` still carries CHEBI:32586, whose *actual* CHEBI label is "sodium sulfate decahydrate" (the wrong compound, with an overridden `label`). The two fields now name different substances. This is explicitly out of this PR's scope (chebi_term canonicalization is deferred) and pre-existing on the MIM side, but the reground makes the intra-record conflict newly visible; worth a follow-up chebi_term reground pass keyed off the same allowlist.

### Findings (could not be anchored — PR diff is truncated server-side, 6,702 files)
- **NIT (recall) — `scripts/fix_primary_term_grounding.py:106`** `basic_search(name)[:12]` caps search hits at 12; a valid single exact match ranked >12 is never seen, so the name falls through to `flag` instead of being grounded. Safe-by-design (under-reaches, never mis-grounds) but shrinks coverage and likely feeds the 1,526 left-untouched pairs. Consider raising the cap or post-filtering exact-label/synonym hits before truncating.
- **NIT (hydrate specificity loss) — allowlist `scripts/fix_primary_term_grounding.py:56`** `Na2SiO3 x 5 H2O` re-grounds CHEBI:86477 (sodium sulfite — correctly a wrong compound) to CHEBI:60720 *sodium silicate* (generic/anhydrous). Net improvement, but pentahydrate specificity is dropped and `term.label` becomes the generic. `preferred_term` keeps the hydrate string, so acceptable.
- **RISK (verify + divergence) — allowlist `scripts/fix_primary_term_grounding.py:61`** `Na2SeO4 x 10 H2O` re-grounds CHEBI:32586 (OBO label *sodium sulfate decahydrate* — the wrong compound) to CHEBI:77775 *sodium selenate*. Primary term now correct, but these records' `mediaingredientmech_chebi_term` still points at CHEBI:32586, so the two fields name different substances (covered under the divergence note above). Decahydrate specificity also lost (acceptable).

realmarcin:
## Conflict resolution + consistency

Merged main; resolved 299 `curation_history` merge conflicts deterministically (kept main's accurate G21/G22/G24 regrounding note plus this branch's primary-term-canonicalization/regrounding events). No `term.label`/`id` content conflicts — the 20,857 canonicalizations didn't overlap main's chebi_term changes. All conflicted files re-validated as parseable YAML.

**Consistency tightened 111 → 109:** the 483 primary-term wrong-id regroundings resolved two reliable-layer same-name→multi-CHEBI ambiguities surfaced by the phase-1 consensus pass. `just check-chebi-grounding` passes at the new, stricter baseline.

PR #51, opened 2026-06-12
Regenerate HTML pages + UMAP visualizations from current data

realmarcin:
## Summary

Full rebuild of CultureMech's published static artifacts so they match the current `normalized_yaml` data + renderer/embeddings. The committed pages were last generated **~2026-05-16**, before the `category`-lowercase migration and the boric-acid `CHEBI:33134`→`CHEBI:33118` fix (#49), so they were broadly stale.

## What was regenerated

| Artifact | Recipe | Notes |
|---|---|---|
| `pages/normalized/` (15,175 files) | `just gen-pages` | lowercase `category` + provenance rows, refreshed `Generated <timestamp>` footer, boric-acid → CHEBI:33118 |
| `app/ingredient_umap.html` | `just gen-ingredient-umap` | 653 ingredients, 512D→2D, `random_state=42`; boric acid regrounded |
| `app/umap.html` | `just gen-media-umap` | 1.1M cached embeddings, 15,827 media, `random_state=42` |

## Notes for review

- **Large diff is expected**: these generators stamp a per-page `Generated <timestamp>` footer, so every page shows a diff even where content is unchanged. The substantive changes are the lowercase-`category` migration and the boric-acid regrounding.
- **UMAPs are deterministically seeded** (`random_state=42`), so layout changes reflect real data deltas, not random churn.
- Verified: `CHEBI:33134` no longer appears anywhere in `pages/` or `app/`.

Part of a cross-repo pass regenerating HTML+UMAP in all four Mech repos.


realmarcin:
Addressed the timestamp-churn issue noted above in commit 14f0c90: `render_media_pages.py` no longer stamps `datetime.now()` in the page/index footer ("Generated {{ generated_at }} from" → "Generated from"). Output is now a pure function of the input YAML — force-re-rendered all 15,827 pages, and a non-force regen now skips unchanged pages (rendered: 0). Future regens won't produce the ~15k-file no-op diff.

PR #88, opened 2026-07-04
Review-cleanup sweep: vendor d3, PaCMAP labels, green tier ramp, data tables, de-emoji

realmarcin:
Applies the remaining dataviz / artifact-design review findings to the CultureMech website. **Stacks on `feat/dark-mode-toggle`** — do not merge before that branch lands. Changes are additive; light and dark themes both preserved.

## Changes
- **R1 — Vendor d3:** copied d3 v7.9.0 to `app/d3.v7.min.js`; replaced the `d3js.org` CDN `<script>` in the `media_umap.html` + `ingredient_umap.html` templates and the generated `app/umap.html`, `app/umap_graph.html`, `app/ingredient_umap.html`.
- **R6 — Marker sizes:** media scatter `r=4` → `r=5`; ingredient radius floor `3` → `5` in `ingredient_umap_generator.py`, plus a `Math.max(5, d.radius)` render guard so the already-generated page renders ≥5px without rerunning the embedding pipeline.
- **Green ordinal tier ramp:** replaced the red/orange/blue rainbow with the validated single-hue green ramp (other `#7cc496`, top500 `#3d9e63`, top100 `#1f6b3f`) in `tierColors`, the legend dots, and the filter-button accents.
- **R2 — PaCMAP relabel:** user-facing copy only (ingredient title/h1/subtitle, `browser.html` nav, `index.html` embedding card). Filenames and data keys (`umap_x/umap_y`, `umap.html`) unchanged. Media map already said PaCMAP; `umap_graph.html` correctly stays "Graph Layout (sfdp)".
- **R3 — Remove decorative emoji:** `browser.html` h1 + nav (🧫 🏠 📊 🔬), `index.html` card icons (🔍 🗺️ 💻), media/ingredient page headers (🧫 🏠) and stat markers (📊 🔍).
- **R5 — Data-table fallback:** keyboard-reachable `<details>` table built from the same data the plot renders, added to `umap.html`, `umap_graph.html`, `ingredient_umap.html`.
- **R4 — Footer self-link:** `index.html` renders the current Mech (CultureMech) as plain text (`aria-current="page"`) instead of a self-link.
- **Discussions toggle:** `app/discussions/index.html` gains a dark theme block, reduced-motion rule, and `../theme-toggle.js`.

Generated pages were edited identically to their `src/culturemech/templates/` sources; the generator's radius floor was bumped for future regenerations.


realmarcin:
Reviewed the incremental diff — clean, approving in spirit (not formally approving/merging).

Verified:
- **Vendored d3**: `app/d3.v7.min.js` is the complete v7.9.0 bundle (279706 bytes, `node --check` passes, terminates correctly). All three plot pages now reference it via the correct relative path `d3.v7.min.js` (they live at app/ root).
- **Green ordinal ramp** `#7cc496 / #3d9e63 / #1f6b3f` applied consistently across all three ingredient-tier surfaces — `tierColors`, legend dots, and filter-button inline styles — with darkest = most-frequent (top100). No leftover rainbow: old `#e74c3c/#e67e22/#3498db` are fully gone (confirmed 0 occurrences in generated `app/umap.html`). Media category colors remain categorical, which is correct for categorical data.
- **Data-table fallback**: field names match the actual data objects — ingredient table (`name/chebi_id/cas_rn/count/tier`) matches the generator's points dict; media table (`entity_type/category/medium_type/physical_state/num_ingredients/num_organisms/source_database`) matches the embedded data. JS var names (`ingredientData`, `derivedData`, `directData`) are correct. All cells HTML-escaped via `esc()` — no unescaped-HTML risk.
- **Template↔generated parity** holds for both umap pages (markers 1:1); radius floor bumped in both the Python generator (max(3→5)) and the JS (`Math.max(5, d.radius)`) — redundant but harmless.
- De-emoji, footer self-link as `<span aria-current="page">`, and PaCMAP relabeling are consistent across pages and templates.

One minor, non-blocking note: the media data-table concatenates derived + direct rows with no column distinguishing the two embeddings, so a medium present in both appears twice. Fine for an accessibility fallback; mentioning only for awareness.

PR #106, opened 2026-07-21
Add role extractor: parse Edison YAML block, emit dual applier batches

realmarcin:
## Summary

Step 7b PR4 of the ingredient-roles migration. Adds `scripts/extract_roles_from_edison.py`, the parser that consumes `research/ingredients/roles/*-edison-literature.md` bundles (produced by PR1's role-research template) and emits two batch JSONs:

- **`--out-mim`** → rich shape with per-role confidence, metabolic_context, and evidence-citation lists. Feeds MIM's `apply_role_research_results.py` (PR2) directly.
- **`--out-cm`** → scalar-projection shape (facet → role-token list, no evidence). Feeds the forthcoming CultureMech `apply_ingredient_roles` (PR5). CultureMech's IngredientDescriptor carries scalar role tokens; evidence stays on the source-of-truth MIM record.

## Robustness notes

- **LAST fenced YAML block wins.** The template's own example lives in an early fence in the response file (the "Question:" block); the model's actual answer lives later. `find_role_yaml_block` iterates in reverse and picks the last block whose top-level key is `role_research:`.
- **Malformed YAML fences skipped, not fatal.** If the model wrote a broken YAML fence somewhere, the extractor moves on to the next candidate rather than aborting.
- **String shorthand upgraded.** `- SULFUR_SOURCE` becomes `{role: SULFUR_SOURCE, confidence: 0.7, evidence: []}`. The model can use either form.
- **Entries without `role:` dropped.** Guards against half-formed proposals.
- **`metabolic_context` schema legality enforced.** Only the cellular-metabolic facet keeps `metabolic_context`; on nutritional/physicochemical it's dropped silently. The applier repeats the check as defense-in-depth.
- **`-citations.md` sidecar cross-referenced.** Evidence entries with a `doi:` or `pmid:` but no `reference_text:` get their reference text filled in from the sidecar. Existing `reference_text:` never overwritten.
- **Meta sidecar fallback.** If the YAML block omits `ingredient:` / `ingredient_identifier:`, the extractor falls back to the `-meta.yaml` sidecar's `slug` / `ingredient_id`.

## Wiring

`just extract-roles-from-edison` (default target: `../MediaIngredientMech/research/ingredients/roles/`).

## Test plan

- [x] `uv run pytest tests/test_extract_roles_from_edison.py` — 24/24 pass
- [x] Rich MIM batch shape verified against PR2 applier's `_load_batch()` — top-level `proposals[]`, each with `ingredient_identifier` / `ingredient_path` / `source_run` / `role_assignments.{nutritional,physicochemical,cellular_metabolic}_roles[]`
- [x] Scalar CM batch shape verified — `roles.{facet}` is a flat list of tokens; no evidence fields leak through
- [x] Real Edison bundles will be exercised end-to-end in PR6 (curator-triggered run against the top-10 prioritizer output from PR3)

### DIFF
--- a/project.justfile
+++ b/project.justfile
@@ -113,6 +113,18 @@ prioritize-deep-research-candidates *args="":
 prioritize-role-research-candidates *args="":
     uv run --extra dev python scripts/prioritize_role_research_candidates.py {{args}}
 
+# Extract structured role assignments from Edison role-research bundles.
+# Consumes `-edison-literature.md` files under a directory (or single files) and
+# emits two batch JSONs:
+#   --out-mim  → rich shape for MIM `apply_role_research_results.py`
+#   --out-cm   → scalar shape for CultureMech's future `apply_ingredient_roles`
+#
+# Default: scans ../MediaIngredientMech/research/ingredients/roles/
+# Example: just extract-roles-from-edison
+[group('Research')]
+extract-roles-from-edison inputs="../MediaIngredientMech/research/ingredients/roles" *args="":
+    uv run --extra dev python scripts/extract_roles_from_edison.py {{inputs}} {{args}}
+
 [group('Research')]
 research-organism-recipe-edison target organism *args="":
     uv run --extra dev python scripts/research_organism_recipe_edison.py \
--- a/scripts/extract_roles_from_edison.py
+++ b/scripts/extract_roles_from_edison.py
@@ -47,7 +47,13 @@ Follows the "never fabricate" rule from the plan: an ingredient whose
 Edison output has no parseable `role_research:` block is skipped and
 recorded in `--skipped-report`; a role whose token is not in the facet's
 enum permissible values (typo, hallucination, wrong facet) is dropped and
-logged. The applier repeats the enum check as a defense-in-depth.
+logged.
+
+This is the ONLY enum check in the pipeline — `apply_ingredient_roles.py`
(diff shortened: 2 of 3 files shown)


realmarcin:
Follow-up commit (f1689da8) tightens the extractor to the applier's exact input contract:

- **Token validation** against the three facet enums, sourced from the vendored `src/culturemech/schema/mim_roles.yaml` (single source of truth; sha-pinned by `verify-schema-pin`). Invalid tokens are DROPPED from the batch, not passed through. Wrong-facet tokens (e.g. `BUFFER` filed under `nutritional_roles`) are dropped AND flagged with the correct facet in `--validation-report`. `--no-validate` disables the check for a backwards-compat safety valve.
- **`source_run`** becomes MIM-repo-relative (per applier contract) when the response file lives inside `--mim-repo` (defaults to `../MediaIngredientMech`); falls back to file stem when not resolvable.
- **`warnings:`** from the YAML block are human-only per the applier contract — they NEVER enter the batch. When `--validation-report` is set, warnings are surfaced there alongside dropped-token records so nothing is silently lost.

Test count is now 35 (was 24), all passing. The 11 new tests cover the schema loader (happy path, missing file, missing enum, sanity-check against the real shipped mim_roles.yaml), token validation (unknown dropped, wrong-facet flagged, permissive-when-no-whitelist), MIM-relative resolution (inside vs outside repo), and CLI validation-report shape.

Issue #116, opened 2026-07-22
Reconcile duplicate media sharing filenames across bacterial/ and archaea/


realmarcin:
Follow-up from #114 / PR #115.

## Problem
Some media exist under **both** `data/normalized_yaml/bacterial/` and `data/normalized_yaml/archaea/` with the **same filename but different content** — so they can't be auto-moved without clobbering.

### The 11 archaeal collisions (excluded from PR #115's move)
All `methano*` (archaeal) media already present in `archaea/` with different content than the `bacterial/` copy:

- marine_methanogenic_medium.yaml
- methanogen_high_salt_medium.yaml
- methanogenium_cv_medium.yaml
- methanogenium_medium.yaml
- methanogenium_medium_h2_co2.yaml
- methanogens_saline_water_medium.yaml
- methanohalophilus_euhalobius_medium.yaml
- methanohalophilus_halophilus_medium.yaml
- methanohalophilus_medium.yaml
- modified_balchs_methanogen_medium_1_b.yaml
- modified_methanohalophilus_medium.yaml

### Broader cross-dir name-collision set (~39 files)
There are also `bacterial/` vs `archaea/` filename collisions that are **correctly** bacterial (halophile/thermophile media that share a name with an archaeal medium), e.g. `thermus_thermophilus_medium.yaml`, `pseudomonas_halophila_medium.yaml`, `halophilic_methanotrophic_bacterium_medium.yaml`. These indicate duplicate imports, not misclassification.

## Ask
For each colliding pair: decide whether it's (a) a true duplicate to dedupe, or (b) two genuinely distinct media that need disambiguated filenames. This is manual curation — the auto-audit deliberately left them alone.

realmarcin:
PR #120 resolves the **archaeal half** of this issue.

**The 11 collisions are not duplicates.** I diffed all 39 cross-directory filename collisions: every pair is two *distinct* media from different source registries — distinct `CultureMech:` ids and distinct compositions (e.g. `methanogenium_medium` is KOMODO/DSMZ-141 in `bacterial/` vs JCM-J265 in `archaea/`). Zero byte-identical pairs. So this issue's question — dedupe or disambiguate — resolves uniformly to **(b) disambiguate**.

Each is renamed with the source prefix the repo already uses for exactly this case (`KOMODO_372_HALOBACTERIA_medium.yaml` and `TOGO_M159_Halobacteria_Medium.yaml` both carry `name: halobacteria_medium`).

[...]

Issue #114, opened 2026-07-22
Sulfolobus & Thermoproteus media mis-filed under bacterial/ (they are archaea)


realmarcin:
## Finding
Two culture media whose target organisms are **archaea** (Crenarchaeota) are stored under `data/normalized_yaml/bacterial/` and are consequently treated as bacterial by downstream tooling:

- `bacterial/sulfolobus_medium_for_dsm_9790.yaml` — *Sulfolobus* (Crenarchaeota)
- `bacterial/TOGO_M1633_Thermoproteus_neutrophilus_Medium.yaml` — *Thermoproteus neutrophilus* (Crenarchaeota)

## How it surfaced
Reviewing the refreshed deep-research triage (`data/import_tracking/reports/deep_research_triage_top10.md`, PR #113), its prose asserts *"All ten are bacterial."* Cross-checking the top-10 against the JSON, #1 (`sulfolobus_medium…`) and #5 (`thermoproteus_neutrophilus_medium`) are archaeal genera. The "bacterial" label traces to the recipes' `file_path` being under `bacterial/` — the prioritizer/triage inherit domain from the directory.

## Impact
- The triage's homogeneity claim ("all bacterial") is inaccurate for 2 of 10.
- Any downstream consumer that infers domain from the `bacterial/` path will mislabel these (and any other archaeal media similarly filed) as Bacteria.

## Scope / fix
1. **Report prose** — corrected in PR #113 (this triage now notes the 8 bacterial + 2 archaeal split and links here). ✅
2. **Root cause (this issue)** — recategorize the mis-filed archaeal media (directory move + any domain field), and audit `bacterial/` for other archaeal genera (e.g. *Sulfolobus*, *Thermoproteus*, *Pyrococcus*, *Methanococcus*, *Halobacterium*, *Thermococcus*). This is a data-curation task beyond the report refresh.

Found during a review of PR #113.

PR #120, opened 2026-07-25
Recategorize 73 more archaeal media; add reusable domain audit (#116)

realmarcin:
Follow-up to #114 / #115, resolving the archaeal half of #116.

## Why there was more to find

#115 moved 629 mis-filed archaeal media using an NCBITaxon **genus-rank label** match. That query had two blind spots:

| Blind spot | Consequence |
|---|---|
| Labels only, no synonyms | `Methanosaeta` is only a *synonym* of `Methanothrix` (NCBITaxon:2222) — every `methanosaeta_*` medium was invisible |
| Genus rank only | `Halobacteria` (class), `Sulfolobales` / `Thermococcales` (order), `Methanobacteriaceae` (family) are not genera — so `halobacteria_*`, `*_sulfolobales_*`, `thermococcales_*` were invisible too |

Matching labels **and** synonyms across genus/family/order/class/phylum, and keeping only names unique to one domain (Archaea vs Bacteria vs Eukaryota), surfaces **75** more records.

## What changed

**73 recipes** move `bacterial/` → `archaea/` with `category: archaea`.
`archaea/` 700 → 773, `bacterial/` 14348 → 14275.

- All 773 archaea files pass `just validate-strict` (0 errors).
- No duplicate `CultureMech:` ids across the tree (15870 unique).
- Re-running the audit is clean, so this is idempotent.

### The 13 filename collisions (#116)

13 of the 73 collided with an existing `archaea/` file. **None are duplicates** — I diffed all 39 cross-directory collisions and every pair is two *distinct* media from different registries (distinct ids, distinct compositions, e.g. KOMODO/DSMZ-141 vs JCM-J265). So #116's question resolves uniformly to "genuinely distinct, needs disambiguation".

Each is renamed with the source prefix the repo **already** uses for this exact case — `KOMODO_372_HALOBACTERIA_medium.yaml` and `TOGO_M159_Halobacteria_Medium.yaml` both carry `name: halobacteria_medium`. Prefixes were verified one-by-one against each record's own `Source:` line.

## Deliberately not fixed here

Both recorded in `data/import_tracking/reports/domain_category_audit.json`:

- **2 MIXED co-culture media** (`*Methanosaeta*` + `*Brevibacterium*`) — genuinely spans both domains. Needs a curator, not a rule; the script never auto-moves these.
- **26 pre-existing `archaea/` records named for bacteria** (`Pseudomonas`, `Thermus`, `Treponema`, `Spirochaeta`, …). I verified via `git log --diff-filter=A` that these predate #115 (commits `ada2845d` / `6cbbb056`) — **#115 introduced no false positives**. This is the remaining, bacterial half of #116.

## Reports intentionally not regenerated

`deep_research_priority*.{json,md}` are now stale for these 73 records, since `category` feeds the score. I regenerated them, saw the top-10 churn wholesale, and reverted:

the prioritizer skips already-researched media by reading the **gitignored** `research/` directory (391 local summaries here). Regenerating on any given machine bakes untracked local state into a committed artifact and reshuffles the ranking for reasons unrelated to this change. That reproducibility problem deserves its own issue rather than a silent piggyback.

## Regression guard

`scripts/audit_domain_categories.py` + `just audit-domain-categories` (report-only by default, `--apply` to move) make this reproducible instead of a one-off sweep.


realmarcin:
## Review

Reviewed the diff (77 files: 73 recipe moves, new audit script, justfile recipe, audit JSON). One data-loss bug, two smaller gaps — all three fixed in 53610d9. Two pre-existing findings filed separately rather than bundled here.

### Blocking — fixed in 53610d9

**Target-name collisions could silently destroy a recipe.** `target_path()` decided a free filename by checking the filesystem, but `main()` computes *every* target before performing *any* move. Two source recipes resolving to the same target both received that name, and the second `git mv` overwrote the first:

```
target 1: TOGO_M1_Foo_Medium.yaml
target 2: TOGO_M1_Foo_Medium.yaml   <- clobbers target 1
```

It did not trigger on this dataset — all 73 targets were distinct, and the counts reconcile exactly (bacterial 14348→14275, archaea 700→773, 15870 unique ids, no duplicates). But the script is committed and `--apply`-able, so it was a latent corruption path in a data-repair tool. Targets are now reserved as handed out, regression-tested against a 3-way pileup:

```
TOGO_M1_Foo_Medium.yaml | TOGO_M1_Foo_Medium_2.yaml | TOGO_M1_Foo_Medium_3.yaml
```

### Non-blocking — also fixed

- `restamp_category()`'s return value was discarded, so a moved file with no `category:` line would land in `archaea/` still claiming `bacterial` to anything reading the field. Now warns on stderr.
- The `Source:` fallback regex scanned the whole document and could pick up an ingredient-level note rather than the record's provenance. Now anchored to the record-level `notes:` line. (Did not affect this PR — all 13 renames derive from `SOURCE_ID_PATTERNS`, and I verified each against its own `Source:` line.)

### Verified

- Audit re-run after the fix is unchanged and idempotent: 0 misfiled, 2 mixed, 26 reverse.
- All 773 `archaea/` files pass `just validate-strict`; CI green (validate / consistency / label-drift).
- All 73 moves are `git mv` renames, so history is preserved.

### Filed separately, not addressed here

- **#124** — 4,774 solution records carry `category: bacterial`, so the prioritizer's documented "category == solutions → dropped" hard filter never fires. They are 31% of the ranked corpus. Pre-existing; the domain axis doesn't really apply to solutions at all.
- **#125** — the committed `*_index.json` files are ~4 months stale and miss ~5,000 recipes. Pre-existing before both #115 and #120.

Neither belongs in a recategorization PR, and #121 (report reproducibility) already covers why the priority reports are deliberately not regenerated here.

Issue #125, opened 2026-07-25
Recipe indexes are ~4 months stale and miss ~5,000 recipes


realmarcin:
## Problem

The committed per-category indexes under `data/normalized_yaml/*_index.json` were all generated **2026-03-16** and have not been refreshed since, while the corpus kept growing. Current drift:

| index | `count` in index | actual `*.yaml` | missing |
|---|---|---|---|
| bacterial | 10,136 | 14,275 | 4,139 |
| archaea | 63 | 773 | 710 |
| specialized | 105 | 455 | 350 |
| algae | 242 | 249 | 7 |
| fungal | 119 | 126 | 7 |

(actual counts as of #120; archaea/bacterial reflect that PR's moves)

Each index entry also records a `filename`, so any recipe that has been renamed or moved between category directories — as #115 and #120 both did in bulk — has a stale path in the index even when the id is present.

## Impact

Any consumer that enumerates recipes via `*_index.json` rather than globbing the directory silently sees roughly a third of the corpus. The indexes are committed artifacts, so this is not obviously visible as staleness — they look authoritative.

## Notes

- `just generate-all-indexes` regenerates them (`scripts/generate_recipe_indexes.py`).
- Neither #115 nor #120 regenerated them; both were already stale beforehand, so this predates both and was deliberately not bundled into either PR.
- Worth considering a CI check that fails when an index disagrees with the directory it describes, so this cannot silently drift again.

Found while reviewing #120.

PR #143, opened 2026-07-29
Decide domain from observed growth, not the medium's name (#138)

### LINKED ISSUE #138: 17 recipes have an unsubstantiated or genuinely ambiguous domain assignment
Residue of the #114 / #116 / #120 / #137 domain work. Everything here is a **judgment call**, which is why the audit deliberately refuses to move it.

## Background

Auditing `archaea/` in both directions showed the original 71-file import (`CultureMech:0002xx-3xx`, DSMZ/JCM) was never domain-vetted: **29 archaeal, 26 bacterial, 16 with no taxonomic evidence at all**. #120 and #137 moved everything decidable by name. This is what's left.

## 15 records in `archaea/` naming no taxon (unsubstantiated)

Their names carry only a *physiology*, and halophiles/thermophiles span both domains — so filing them under `archaea/` is a claim with nothing behind it. They may well be archaeal; nothing in the record says so.

```
DSMZ_1399_HALOPHILIC_MEDIUM.yaml            HALOPHILIC MEDIUM
JCM_J1334_NATRONOARCHAEA_MEDIUM_II.yaml     NATRONOARCHAEA MEDIUM II
JCM_J464_HP_101_HALOPHILE_MEDIUM.yaml       HP 101 HALOPHILE MEDIUM
acido_thermophile_medium.yaml               ACIDO-THERMOPHILE MEDIUM
alkalophilic_halophile_medium.yaml          ALKALOPHILIC HALOPHILE MEDIUM
artificial_seawater_for_halophiles.yaml     ARTIFICIAL SEAWATER FOR HALOPHILES
halophile_medium.yaml                       HALOPHILE MEDIUM
halophile_starch_casein_medium.yaml         HALOPHILE STARCH-CASEIN MEDIUM
halophilic_medium.yaml                      HALOPHILIC MEDIUM
halophilic_sob_medium.yaml                  HALOPHILIC SOB MEDIUM
hp_101_halophile_medium.yaml                HP 101 HALOPHILE MEDIUM
medium_for_aciduric_thermophilic_bacillus_strains.yaml
moderate_halophile_medium.yaml              MODERATE HALOPHILE MEDIUM
moderate_halophile_medium_2.yaml            MODERATE HALOPHILE MEDIUM 2
thermanaerobacter_thermophila_medium.yaml   THERMANAEROBACTER THERMOPHILA MEDIUM
```

Three look resolvable from the name alone but were left for a human on purpose:

- `JCM_J1334_NATRONOARCHAEA_MEDIUM_II` — "Natronoarchaea" is not an NCBITaxon name, so the audit can't confirm it; it is almost certainly archaeal and already sits in `archaea/`.
- `medium_for_aciduric_thermophilic_bacillus_strains` and `thermanaerobacter_thermophila_medium` — *Bacillus* is a cross-domain homonym (also a diatom genus) with no binomial to disambiguate, and *Thermanaerobacter* looks like a misspelling of *Thermoanaerobacter*. Both read as **bacterial** and probably belong in `bacterial/`, but that's a curator's call, not a rule's.

## 2 genuine co-culture media

```
bacterial/TOGO_M478_Methanosaeta_Brevibacterium_Medium.yaml
bacterial/methanosaeta_brevibacterium_medium.yaml
```

Name an archaeon *and* a bacterium. The single-valued `category` field can't express this — arguably the real fix is schema-level rather than picking a winner.

## Reproduce

```
just audit-domain-categories
```
Machine-readable lists live in `data/import_tracking/reports/domain_category_audit.json` under `unresolved` and `mixed`.

## Note on #116

#116 is closed, but PR #134 (its most recent work) explicitly said *"This does not close #116"* — it was a read-only triage. #137 completes the bacterial half. This issue tracks the genuinely-undecidable remainder so it isn't lost.

realmarcin:
## #138 is valid — reproduced exactly (15 unresolved + 2 mixed)

Its residue is 15 records in `archaea/` whose names carry only a *physiology* — "HALOPHILE MEDIUM", "ACIDO-THERMOPHILE MEDIUM". Halophiles and thermophiles span both domains, so the placement was an assertion with nothing behind it. The name-based audit was right to refuse to guess.

## A second, independent evidence source

kg-microbe's MediaDive transform records which taxa were *observed* to grow in each medium (`METPO:2000517`). Resolving those to a domain answers the question from observation rather than the label:

| | |
|--:|---|
| **4** | confirmed archaeal — *Halogeometricum*, *Halomicrobium*, *Halorubrum vacuolatum*, *Haloquadratum walsbyi*, *Haladaptatus paucihalophilus* |
| **6** | shown **bacterial** — *Halobacillus*, *Virgibacillus*, *Acidothermus cellulolyticus*, *Rhodovibrio salinarum*, *Pseudomonas*, *Salinicoccus* |
| **5** | no growth evidence — still judgment calls, untouched |

The evidence is decisive exactly where the name misleads: **Halobacillus** and **Virgibacillus** are Bacillota despite halophile naming, while **Halorubrum** and **Haladaptatus** are Halobacteriales. No name heuristic separates those.

## Moved 3 of the 6 — and deliberately not the other 3

```
moved   JCM_J464_HP_101_HALOPHILE_MEDIUM
moved   acido_thermophile_medium
moved   moderate_halophile_medium -> JCM_J599_moderate_halophile_medium
```

The rename is because a **different** medium already holds that filename in `bacterial/` (9/11 ingredients shared, distinct ids) — so the source id disambiguates rather than clobbers (#116).

**Not moved:** `halophile_medium`, `hp_101_halophile_medium`, `medium_for_aciduric_thermophilic_bacillus_strains`. Each is a **verbatim duplicate** of a record already in `bacterial/` (`CultureMech:006229` / `:006131` / `:006270` — same ingredients, groundings, concentrations and units, per the #134 classifier). Moving them would put two identical records in one directory under different names, which is worse than the status quo. Deduping means deleting a CultureMech id — a curation decision, not an audit's.

## Notes

`scripts/domain_growth_evidence.py` degrades to "no evidence" when kg-microbe is absent, so **CI behaviour is unchanged** — no runner has that checkout.

A test caught a real bug in my own module: `load_growth_edges` kept a trailing newline on the object column. Harmless against the live 9-column export, silently match-nothing on a narrower one.

**The #125 index guard failed on these moves before I did** — first time it has caught a live change rather than pre-existing drift. Indexes regenerated.

## Verification

- [x] 16 new tests, incl. cyclic-lineage termination and "refuses to decide when both domains present"
- [x] `validate-strict data/normalized_yaml/bacterial` — 14,305 files, 0 errors
- [x] Suite: 514 passed, 2 skipped
- [x] Audit `unresolved` 15 → 12; `archaea/` 773 → 770, `bacterial/` 14,275 → 14,278

**Does not close #138** — 12 unresolved + the 2 co-culture records remain, and the co-culture case is arguably schema-level (single-valued `category` can't express it).


realmarcin:
## Review of this PR — 3 findings, 2 addressed here

### 1. Moving records left stale paths in a tracked report — **fixed**

`data/import_tracking/reports/deep_research_priority.json` carried three `archaea/…` paths that no longer exist. Regenerated; **0 stale entries** remain.

The 145-line churn is not noise: the three moved records changed category, so their multiplier went `archaea 0.95 → bacterial 1.00`, which reorders neighbouring scores. Expected consequence of the move, not formatting.

### 2. The id registry — **my 3 fixed; a much larger problem found**

`data/culturemech_id_registry.tsv` is tracked and maps id → path. This PR made 3 rows stale and they are now repaired.

But an honest correction to my first read of it: the registry has **5,511 stale rows that predate this PR**, left behind by the bulk moves in #115 / #120 / #137. I initially reported "3 stale rows"; the real number is three orders of magnitude larger, and this PR is a rounding error on it rather than the cause.

Filed as **#144**, including the guard that would have caught all of them — the direct analogue of `tests/test_recipe_indexes.py`, which caught *this* PR's index staleness within seconds of the move.

### 3. Only one derived artifact is guarded — **filed as #145, not fixed**

| artifact | tracked | guarded |
|---|---|---|
| `*_index.json` | yes | **yes** (#125) |
| `culturemech_id_registry.tsv` | yes | no |
| `deep_research_priority*.json` | yes | no |
| `reports/*.tsv`, `*.json` | yes | no |

Worth noting the `reports/` case cuts the other way: files like `label_plausibility_2026-07-19.tsv` are **dated snapshots**, and their stale paths are *correct* — they record what was true when they ran. Refreshing them would falsify history. So the fix is not "regenerate everything", it is classifying each artifact as current-view or snapshot. That distinction isn't written down anywhere, which is why it is easy to get wrong in both directions.

### Verification

- `just validate` on all three moved records — passes
- `validate-strict data/normalized_yaml/bacterial` — 14,305 files, 0 errors
- Suite — 514 passed, 2 skipped
- 0 stale paths remain in the priority report; registry rows for these 3 resolve

### What I did not change

The 3 verbatim-duplicate records (`halophile_medium`, `hp_101_halophile_medium`, `medium_for_aciduric_thermophilic_bacillus_strains`) are still in `archaea/` despite growth evidence saying bacterial — moving them would put two identical records in one directory, and deduping deletes a CultureMech id. Still a curation call.

Issue #144, opened 2026-07-29
culturemech_id_registry.tsv has 5,511 stale rows — id→path lookups resolve to files that no longer exist


realmarcin:
## Problem

`data/culturemech_id_registry.tsv` maps every CultureMech id to its file path. It is tracked, and **5,511 of its rows point at files that do not exist**:

```
CultureMech:000243  data/normalized_yaml/archaea/thioalkalivibrio_halophilus_medium.yaml
CultureMech:000244  data/normalized_yaml/archaea/thiohalophilus_medium.yaml
CultureMech:000245  data/normalized_yaml/archaea/thiohalospira_halophila_medium.yaml
CultureMech:000248  data/normalized_yaml/archaea/halophilic_methanotrophic_bacterium_medium.yaml
...
```

Reproduce:

```python
from pathlib import Path
bad = [l for i, l in enumerate(Path("data/culturemech_id_registry.tsv").read_text().splitlines())
       if i and l.strip() and not Path(l.partition("\t")[2]).is_file()]
len(bad)  # 5511
```

## Cause

Bulk category moves. #115 moved 629 records, #120 another 73, #137 more, and #143 three — every one of those changed a file's path, and none updated the registry. The rows are not wrong about the *id*, only about where the record now lives.

## Impact

Anything resolving an id through the registry rather than by scanning the corpus gets a path that does not open. This is the same class of defect as #125 (stale indexes) but on a different artifact — and unlike the indexes, nothing guards it.

## Suggested fix

1. Refresh the registry from the corpus. `scripts/assign_culturemech_ids.py` writes it, but it also *assigns* ids, so it needs checking that a refresh mode exists which only rewrites paths and never mints or renumbers.
2. Add a guard alongside `tests/test_recipe_indexes.py` asserting every registry row resolves to an existing file. That test caught #143's index staleness immediately; the equivalent here would have caught all of the above.
3. Consider whether path belongs in the registry at all. The id is stable, the path is not — a registry keyed on id that stores a *category* rather than a full path would not rot on every move.

## Note

#143 contributed 3 of these and fixed its own 3, leaving 5,511 that predate it. Fixing the rest is a corpus-wide job rather than something to bundle into a domain-audit PR.

PR #159, opened 2026-07-31
Remove the dead taxonomy classifier rather than rewrite it (closes #154)

### LINKED ISSUE #154: TaxonomyClassifier branches on deprecated medium_type values that appear on zero records
## `TaxonomyClassifier.classify_functional_domain` reads a slot that is now deprecated, using values that never existed

`src/culturemech/taxonomy/classifier.py:90-103` branches on `medium_type`:

```python
medium_type = recipe.get('medium_type', '')
if medium_type == 'MINIMAL':      return 'MINIMAL'
elif medium_type == 'COMPLEX':    return 'COMPLEX'
elif medium_type == 'SELECTIVE':  return 'SELECTIVE'
elif medium_type == 'DIFFERENTIAL': return 'DIFFERENTIAL'
elif medium_type == 'ENRICHMENT': return 'ENRICHMENT'
```

Two problems, one pre-existing and one introduced by #148.

**Four of those five branches are unreachable and always were.** The corpus has
only ever held two `medium_type` values:

```
8612 COMPLEX
2480 DEFINED
   1 NEGATIVE_CONTROL
   1 BUFFER
```

`MINIMAL`, `SELECTIVE`, `DIFFERENTIAL` and `ENRICHMENT` appear on zero records,
so only the `COMPLEX` branch has ever fired. Records fall through to the
`applications` keyword heuristics below instead.

**And the slot is now deprecated.** After #148 the live signal is
`composition_type` / `nutritional_class` / `functional_role`. This function
still reads the old slot, so it will not see anything a curator sets on the new
axes.

## Severity is low

`classifier.py` sits at **7% coverage** and `classify_functional_domain` has no
call site outside the module — `taxonomy/__init__.py` exports the class but
nothing in the pipeline, scripts, or tests invokes it. So this is stale code
rather than a live wrong answer.

Worth either rewriting against the three axes or deleting, but it should be a
deliberate decision rather than a drive-by change inside #148.

realmarcin:
## #154's facts check out

```
8612 COMPLEX   2480 DEFINED   1 NEGATIVE_CONTROL   1 BUFFER
```

`MINIMAL` / `SELECTIVE` / `DIFFERENTIAL` / `ENRICHMENT` are on **zero** records, and `culturemech.yaml` now marks `medium_type` `deprecated:` after #148.

## Two corrections, both of which change the remedy

**1. It is not unreferenced.** The issue says nothing invokes the class. `scripts/assign_taxonomy.py:45` instantiates it and `yaml.dump`s the result back over the record (line 106). It is reachable, and it writes to the corpus.

**2. But it could never have worked.** The block it writes — `taxonomy:` — has no schema class. `MediaTaxonomy` and `FunctionalDomainEnum` appear nowhere in `culturemech.yaml` **and nowhere in its git history**. They were never written, not removed.

Probed it directly, adding the exact block the classifier produces to a clean record:

```
[ERROR] Additional properties are not allowed ('taxonomy' was unexpected) in /
```

So the script would have written schema-invalid records corpus-wide. That is why no record carries a `taxonomy:` block — not because nothing calls it, but because it has never been runnable. **Unfinished work, not decayed work.**

So the severity reading lands in the same place as the issue's, by a different route: no bad data exists. But it's a loaded gun rather than a disconnected one — one `python scripts/assign_taxonomy.py` away from 11,092 invalid records.

## Why delete rather than rewrite

Rewriting means first **designing the missing schema class** against the three axes #148 introduced, because the old single `domain` conflated all three:

| old `domain` value | now lives on |
|---|---|
| `COMPLEX` | `composition_type` → `UNDEFINED` |
| `MINIMAL` | `nutritional_class` |
| `SELECTIVE` / `DIFFERENTIAL` / `ENRICHMENT` | `functional_role` (multivalued) |

That is a feature to specify, not a repair to make — and leaving a non-functional implementation in the tree while it's specified only invites someone to run it.

`SimilarityCalculator` and `UnitConverter` are untouched; they have tests and are unrelated.

## Docs

`docs/TAXONOMY_SYSTEM.md` described all of this as working software. Rather than delete 289 lines of design thinking, it now carries a status banner recording what was never built and why, with the usage examples for the removed entry points neutralised. Leaving docs pointing at deleted code is the same staleness class as #125 and #144.

## Test plan

- [x] Every claim in the issue independently verified before acting
- [x] Schema absence confirmed against git history, not just the current file
- [x] Invalidity proven by probe, not inferred
- [x] No remaining reference to the removed names in code, justfile or docs
- [x] `culturemech.taxonomy` still imports; `TaxonomyClassifier` correctly gone
- [x] Suite: 547 passed, 2 skipped

Closes #154.


realmarcin:
Closing in favour of `chore/autonomy-groundwork`, which carries this change as commit `c9b46698`.

Verified there is nothing unique here to lose:

| file | status |
|---|---|
| `docs/TAXONOMY_SYSTEM.md` | **byte-identical** blob in both |
| `src/culturemech/taxonomy/__init__.py` | **byte-identical** blob in both |
| `scripts/assign_taxonomy.py` | deleted in both |
| `src/culturemech/taxonomy/classifier.py` | deleted in both |

The only divergence was `NEXT_TASKS.md`, and it is **superseded rather than lost**: 24 lines from this branch are absent there because `d69ef918` ("correct the stale validate-media-recipes warning") removed those exact lines on purpose. Keeping them would reintroduce a warning that branch had just corrected.

Those `NEXT_TASKS.md` lines should not have been in this PR at all — they were another session's uncommitted working-tree changes, swept in by `git add -A` on a shared checkout. My error; no harm done, since the surviving branch supersedes them.

The investigation behind the change is preserved in `c9b46698`'s commit message and in #154. Its substance, for the record: the classifier was not merely stale but **unrunnable** — the `taxonomy:` block it wrote has no schema class (`MediaTaxonomy` and `FunctionalDomainEnum` appear nowhere in `culturemech.yaml` or its git history), so its output fails validation with `Additional properties are not allowed ('taxonomy' was unexpected)`. That is why deletion was chosen over a rewrite.

Issue #150, opened 2026-07-31
Corpus repair for the implausible-concentration audit (unfinished half of #118)


realmarcin:
Splitting the unfinished half of #118 into its own issue.

#118 asked for two things: an audit to find where stock-solution concentrations and unit slips were stored as final per-litre values, and the corpus repair. **#135 shipped the audit only** — its merge commit says so outright:

> `Does NOT close #118; the corpus repair remains.`

#118 was nonetheless closed on 2026-07-26. The corpus is still wrong; what changed is that we now know exactly where. This issue tracks the repair.

## What the audit found

Three read-only detectors, output in `data/import_tracking/reports/concentration_plausibility.tsv` and `…_by_record.tsv`. Re-counted from the committed reports on 2026-07-30:

| detector | rows | signal |
|---|--:|---|
| `INDICATOR_UNIT_SLIP` | 5,601 | indicators/vitamins ≥ 0.1 G_PER_L — a ~1000x slip |
| `TRACE_SALT_AS_STOCK` | 5,341 | trace-element salts ≥ 1 G_PER_L — stock-solution magnitude |
| `WATER_AS_VOLUME` | 598 | water ≥ 1000 G_PER_L — a prep volume flattened into the ingredient list |

**11,540 rows across 3,914 records.** All three records named in #118 are detected and pinned by a regression test.

## The actionable subset

**579 records are flagged `flattened_cocktail`** — three or more flagged vitamin/trace rows *and* no `solutions:` block. For these the repair is structural ("nest this cocktail under a solution object with an addition volume") rather than per-value, which makes them both higher-value and more mechanical than the rest. The remaining 3,335 flagged records are not cocktails and need per-value judgment.

`DSMZ_962a_THERMOVENABULUM_MEDIUM` is the worked example: the entire DSMZ vitamin solution flattened into `ingredients:` at stock strength.

## Scope limits — read before trusting the counts

- **Only `G_PER_L` rows are examined** (149,289 of 166,684). `MILLIMOLAR`, `MG_PER_L` and `VARIABLE` are untouched. A molar-basis check needs molecular weights and is separate work.
- **Detectors key on confirmed failure modes, not outlier statistics**, so bulk-nutrient errors are missed. `MARICAULIS_medium` carries `Peptone 100 G_PER_L` and `Yeast extract 50 G_PER_L`, both implausible, neither flagged.
[...]

Issue #145, opened 2026-07-29
Moving a record silently invalidates several tracked derived artifacts; only indexes are guarded


realmarcin:
## Problem

A category move (`git mv` + restamp `category:`) leaves a record's *old* path embedded in several tracked artifacts. There is no single step that refreshes them, and no guard on most of them.

Observed after #143 moved three records:

| artifact | tracked | guarded | refreshed by |
|---|---|---|---|
| `data/normalized_yaml/*_index.json` | yes | **yes** — `tests/test_recipe_indexes.py` (#125) | `just generate-indexes` |
| `data/culturemech_id_registry.tsv` | yes | no | `scripts/assign_culturemech_ids.py` (see #144) |
| `data/import_tracking/reports/deep_research_priority*.json` | yes | no | `just prioritize-deep-research-candidates` |
| `reports/*.tsv`, `reports/*.json` | yes | no | various one-off scripts |

Only the first fails loudly. #143's index staleness was caught by its test within seconds; the registry and priority-report staleness were found only because the author went looking.

## The `reports/` case is different

Files like `reports/label_plausibility_2026-07-19.tsv` are **dated snapshots**. Their stale paths are arguably correct — they record what was true when they ran. Rewriting them would falsify history. So the fix is not "regenerate everything"; it is deciding, per artifact, whether it is a *current view* (must refresh) or a *snapshot* (must not).

That distinction is not currently expressed anywhere, which is why it is easy to get wrong in either direction.

## Suggested fix

1. Classify each tracked derived artifact as current-view or snapshot, and record it — a short table in `docs/DATA_LAYERS.md` would do.
2. For current-view artifacts, add freshness guards modelled on `tests/test_recipe_indexes.py`. That test has already proved it works.
3. Consider a `just refresh-derived` recipe that regenerates exactly the current-view set, so a bulk move has one obvious follow-up step instead of a checklist held in someone's head.

Found while reviewing #143.

realmarcin:
Tally of this umbrella after the #148 review — four instances found, three now closed.

| artifact | issue | state |
|---|---|---|
| `*_index.json` recipe indexes | #125 | closed by #133 |
| `data/culturemech_id_registry.tsv` | #144 | closed by #147 |
| `culturemech_dataclasses.py` | #149 | closed by #148 |
| `reports/chebi_consistency.tsv` | #157 | **open** |

[...]

PR #160, opened 2026-07-31
Groundwork: CI-portable justfile, SKILL.md normalisation, remove dead taxonomy classifier

### LINKED ISSUE #154: TaxonomyClassifier branches on deprecated medium_type values that appear on zero records

realmarcin:
Four commits of infrastructure groundwork. Each is independently reversible; grouped because they are all "things that only worked on one machine".

## `Make justfile recipes CI-portable` (8e8dba5)

Eleven recipes hardcoded `/opt/homebrew/bin/python3.13`, absent on any CI runner. Replaced with `uv run python`, matching the convention the rest of the justfile already used. Two scripts carried the same path in their shebang.

Note the reasoning against the obvious alternative: `python3` would have been **wrong**, because on the dev machine that resolves to miniforge 3.13.12 rather than Homebrew 3.13.8 — different site-packages, so the swap would have silently changed dependency resolution.

`matplotlib` is now declared. `kg_microbe_qc` imports it unconditionally, so `gen-qc-dashboard` only ever worked because Homebrew's python happened to have it globally.

## `Normalise skill filenames to SKILL.md` (ddfc1e2)

4 `SKILL.md` against 10 `skill.md`. macOS is case-insensitive (`core.ignorecase=true`) so the split was invisible locally; **Linux CI is not**, and an agent running there would silently fail to find its own instructions.

## `Remove the dead taxonomy classifier` (c9b4669) — closes #154

`TaxonomyClassifier` wrote a `taxonomy:` block whose schema class was **never written** — `MediaTaxonomy` and `FunctionalDomainEnum` appear nowhere in `culturemech.yaml` or its git history. Probing a record with the block it produces gives `Additional properties are not allowed ('taxonomy' was unexpected)`.

So it was unrunnable rather than merely stale, which is what settled delete-over-rewrite: reviving it means first designing the missing schema class against the three axes #148 introduced.

Contrary to #154's text, it *was* reachable — `scripts/assign_taxonomy.py` instantiates it and writes records. A loaded gun, not a disconnected one.

## `NEXT_TASKS` reconciliation (d69ef91)

Notes #150/#151 and corrects a stale `validate-media-recipes` warning.

---

Review, findings and verification follow in a comment.


realmarcin:
## Review — 1 finding, addressed here; verified the rest

### Finding: the SKILL.md rename left three links behind — **fixed**

`.claude/skills/review-recipes/SKILL.md` still pointed at:

```
../create-recipe/skill.md
../manage-identifiers/skill.md
../manage-ingredient-hierarchy/skill.md
```

The rename commit deliberately fixed the equivalents in `review-recipes/reference/` and missed these. Both spellings open on macOS, so nothing local could show the difference — **the same root cause as the split it was fixing**.

Added `tests/test_markdown_links.py` so this cannot recur. `Path.exists()` is unusable here: on a case-insensitive filesystem it returns `True` for a wrong-case path, so the check compares the filename against an actual directory listing.

### The guard found a second, unrelated break on its first run

`docs/MAPPING_METHOD_INDICATOR.md` linked to `NORMALIZATION_IMPROVEMENTS.md`, which had been moved to `docs/archive/`. Nothing to do with the rename; repointed.

That find changed the test's design — it now reports two failure modes **separately**, because their causes differ:

- **wrong case** — works locally, breaks only on Linux. Insidious; the reason this test exists.
- **missing target** — broken everywhere, usually means the file moved.

The original single message would have mislabelled the moved-file case as a case-sensitivity problem.

### Also refreshed

`data/import_tracking/reports/filename_collisions.tsv` — the review incidentally showed it stale. 16 listed collisions no longer exist, resolved by #137/#148. Verified each individually (archaea copies gone, bacterial retained). Leaving a knowingly-wrong tracked artifact to keep the diff tidy is the #145 failure mode.

### Verified, not taken on trust

| claim | result |
|---|---|
| no hardcoded `/opt/homebrew` paths remain | 0 hits across justfile, scripts, src, workflows |
| no lowercase `skill.md` remains | 0 tracked; 14 `SKILL.md` |
| recipes actually run | `validate`, `audit-filename-collisions`, `refresh-id-registry --dry-run` all OK |
| `matplotlib` declared | `pyproject.toml:53` |
| `_require-claw` fails loudly | confirmed — exits 1 with an actionable message |
| taxonomy removal complete | `culturemech.taxonomy` imports; `TaxonomyClassifier` gone; no dangling refs |

The `python3`-would-have-been-wrong reasoning in 8e8dba5 is worth preserving — miniforge 3.13.12 vs Homebrew 3.13.8 have different site-packages, so the obvious swap would have silently changed dependency resolution.

Suite: **803 passed, 2 skipped**.

PR #163, opened 2026-08-01
Add the append-only curation-history layer

realmarcin:
Ports the curation-history layer from TraitMech, which has been running it in production. Same four-piece surface, adapted to CultureMech's paths and data layout.

## Why

Nothing here records *which model, using which tool, changed what, why, and under which issue*. Git tells you a commit happened; it does not tell you an ingredient was grounded against a specific CHEBI release, which deep-research provider produced a recipe correction, or that a review deliberately changed nothing. That gap widens as autonomous agents do more of the changing.

The directory-per-slug plus unguessable shortid layout means two agents curating the same medium concurrently cannot write the same file — this layer has **no merge-conflict surface**. That matters more here than in most repos: CultureMech carries over 15,000 media records and several agents work it at once.

## What's in it

- **`src/culturemech/schema/history.yaml`** — vendored byte-identical from claw's canonical `shared/history/history.yaml` (md5 `3742bc2068b637868c48aba406f6569d`, matching both the canonical copy and TraitMech's vendored one). Vendored rather than referenced because claw is private: public CI cannot check it out without a token. `just validate-history` and the workflow both use this local copy and work with no claw checkout at all. Only the `new-history` scaffolder needs claw, and that is a dev-time tool.
- **Two justfile recipes.** `new-history` resolves the scaffolder through `CLAW_SRC` and fails loudly when it is missing — no skip path, since a skip-when-missing check is exactly what lets a job pass while verifying nothing. `validate-history` points at the vendored schema.
- **`history/README.md`** — the vocabulary, the layout rationale, and a runnable `diff` for checking vendored-schema drift (deliberately not an md5 written into prose, which goes stale silently).
- **`.github/workflows/curation-history.yaml`** — deliberately asymmetric. Record **validity is blocking**; the **missing-record presence check is advisory**, warning in the job summary and never failing. A hard gate on provenance blocks legitimate work at inconvenient moments and trains people to route around it. Path filters point at `data/normalized_yaml/**`, the hand-curated surface that `validate-strict` also guards.

## `set positional-arguments := true`

Required for `new-history`: `--summary` and `--details` are prose, and plain `{{args}}` interpolation splits them on whitespace.

I checked before enabling it — **no recipe in `justfile` or `project.justfile` uses `$1`, `$@`, `$2`, or `$*`**, so nothing can regress. Confirmed empirically as well: `just -n` output for the existing `*args` recipes is byte-identical before and after the change.

```
$ just -n migrate-medium-type-axes --apply
uv run --extra dev python scripts/migrate_medium_type_axes.py --apply     # identical pre/post
$ just -n validate-media-recipe 1_10_r2a_medium --out /tmp/x
uv run python scripts/render_media_prompt.py --target 1_10_r2a_medium --out /tmp/x   # identical pre/post
```

## Verification

| Check | Result |
|---|---|
| `just --list` recipe count | 211 → 213, exactly the two additions |
| Vendored schema md5 vs claw canonical | identical |
| Scaffold + validate a real record | `No issues found`, exit 0 |
| Scaffold **without** `--details` | `validate-history` **fails** on the schema's TODO-placeholder pattern, exit 1 |
| Multi-word `--summary` through `just` | survives intact, unsplit |
| `CLAW_SRC=/nonexistent just new-history …` | fails loudly, exit 1 |
| `just validate-history ""` | exit 2 |
| `just validate-history <nonexistent>` | exit 2 |
| Empty history directory | exit 0, "No history records under …" |

The TODO-placeholder failure is the important one — it means an unfilled record cannot slip through.

The one included history record describes this port itself, and no media records were touched.

## Did not translate cleanly

TraitMech's README closes by noting its vendored schema is not on the automated vendored-fleet drift check, tracked in its #191. CultureMech has no vendored-drift check at all to append this to, so the README states the gap and gives the manual `diff` instead of citing an issue number I would have had to invent.

### DIFF
--- a/.github/workflows/curation-history.yaml
+++ b/.github/workflows/curation-history.yaml
@@ -1,0 +1,94 @@
+name: Curation history
+
+# Two-part check, deliberately asymmetric (see history/README.md):
+#
+#   * VALIDITY is blocking. A malformed history record fails like any other
+#     validation error.
+#   * PRESENCE is advisory. If a media recipe changes without a matching history
+#     record, this warns in the job summary and passes. A hard gate on provenance
+#     blocks legitimate work at inconvenient moments and trains people to route
+#     around it.
+#
+# Validation uses the VENDORED schema at src/culturemech/schema/history.yaml, so
+# this job has no dependency on the private culturebotai-claw repo. Only the
+# `just new-history` scaffolder needs claw, and that is a dev-time tool.
+
+on:
+  pull_request:
+    paths:
+      - "data/normalized_yaml/**"
+      - "history/**"
+      - "src/culturemech/schema/history.yaml"
+      - ".github/workflows/curation-history.yaml"
+  push:
+    branches: [main]
+    paths:
+      - "data/normalized_yaml/**"
+      - "history/**"
+      - "src/culturemech/schema/history.yaml"
+      - ".github/workflows/curation-history.yaml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: curation-history-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  history:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0   # need the merge base to diff against
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+
+      - name: Install deps
+        run: uv sync
+
+      - name: Validate history records (blocking)
+        run: |
+          set -euo pipefail
+          if [ ! -d history ] || [ -z "$(find history -name '*.yaml' -print -quit)" ]; then
+            echo "No history records to validate."
+            exit 0
+          fi
+          find history -name '*.yaml' -print0 \
+            | xargs -0 uv run linkml-validate \
+                --schema src/culturemech/schema/history.yaml \
+                --target-class HistoryRecord
+
+      - name: Check for missing history records (advisory)
+        if: github.event_name == 'pull_request'
+        run: |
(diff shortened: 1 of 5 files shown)


realmarcin:
## Review

Reviewed against the TraitMech reference and against the two sibling ports, since three agents wrote these independently and divergence between them is itself a signal.

**Clean on every load-bearing property:**

- Vendored `src/culturemech/schema/history.yaml` is **byte-identical** to claw's canonical copy (md5 `3742bc2068b637868c48aba406f6569d`, verified across all five copies in the fleet). The whole design rests on this — spokes validate with plain `linkml-validate` against a local copy and need no claw checkout.
- Record validation is **blocking**; the missing-record presence check is **advisory** and cannot exit non-zero.
- `new-history` **fails loudly** when `CLAW_SRC` is missing. No skip-and-exit-0 path — that failure mode is why TraitMech#182 existed.
- `fetch-depth: 0` with an explicit `git fetch` before the `merge-base` diff, so the advisory diff actually resolves.
- `set positional-arguments := true` was needed and is safe: no pre-existing recipe in `justfile` or `project.justfile` uses `$1`/`$@`.
- `data/normalized_yaml/**/*.yaml` matches 15,878 tracked files. Worth noting this is correct **only because that tree is genuinely nested** — MIM's port used the same `**` form against a flat directory and matched zero files (MIM#164). Not a latent problem here.
- README claims check out: the cited `--target-root data/normalized_yaml/bacterial` exists with 14,305 records, the cited schema path exists, and the claim that the TODO placeholder is rejected is backed by the pattern actually being present in the vendored schema.

**The committed history record is accurate**, not fabricated. It cites the md5, names what was verified, and records that `set positional-arguments` was checked safe beforehand. That is the layer doing its job on its first real use.

No issues to file. All four checks green. Merging.

PR #164, opened 2026-08-02
Restamp 239 self-contradicting DEFINED records; guard the invariant (#158)

### LINKED ISSUE #158: 284 DEFINED records (11.5%) list chemically undefined ingredients
## 284 records assert `composition_type: DEFINED` while listing chemically undefined ingredients

`MediumCompositionTypeEnum.DEFINED` means "every component and its exact quantity
is known". 11.5% of the records carrying it contradict that in their own
ingredient list.

Scanning `preferred_term` on every ingredient for undefined components (yeast
extract, peptone, tryptone, casamino acids, beef/meat/malt extract, proteose,
brain-heart infusion, casein hydrolysate, soytone, blood, serum, rumen fluid,
liver extract, tryptic soy):

```
DEFINED   : 2480
  containing an undefined component : 284  (11.5%)
UNDEFINED : 8612
  exactly one undefined component   : 3116
  two or more                       : 3716
  none recognized                   : 1780
```

Examples, none of them borderline:

```
medium_465_modified_for_dsm_10063.yaml       -> casamino, meat extract, peptone, proteose, yeast extract  (35 ingredients)
for_dsm_13514.yaml                           -> peptone, yeast extract       (17 ingredients)
medium_457_modified_for_dsm_11565.yaml       -> meat extract, peptone        (28 ingredients)
MEDIADB_460_Defined_freshwater_medium_CoCl2.yaml -> tryptone, yeast extract  (31 ingredients)
```

The `medium_457_modified_for_dsm_*` family is a large cluster of near-identical
records that all carry the same contradiction, so the 284 are concentrated
rather than uniformly scattered — a modest number of source recipes probably
accounts for most of them.

## Not introduced by #148

The migration mapped `medium_type: DEFINED` → `composition_type: DEFINED`
faithfully and never invented a value. The error was already present in
`medium_type`; #148 restated it on a slot whose description is explicit about
what DEFINED means, which is what makes it detectable now. Note
`MEDIADB_460_Defined_freshwater_medium_CoCl2` has "Defined" in its *source
name*, so at least some of these came in mislabeled from upstream.

## Suggested handling

Most of these are probably `SEMI_DEFINED` rather than `UNDEFINED` — a defined
base plus a named undefined supplement is exactly that case, and `SEMI_DEFINED`
is currently emitted on zero records (#152).

Worth doing as a detector plus a review pass rather than a blind flip, since the
keyword list above will have both false positives (an ingredient named "peptone
water" used as a diluent) and false negatives (undefined components not on the
list). The detector belongs next to the implausible-concentration checks from
#118/#135.

Found while reviewing #148. Companion to #152, which covers the axes that ship
empty; this one covers a value that ships wrong.

realmarcin:
## The contradiction

`MediumCompositionTypeEnum.DEFINED` means *"every component and its exact quantity is known"*. **285 records** assert it while listing yeast extract, peptone, tryptone or meat extract — chemically undefined by definition.

Verified independently before acting. My count is 285 against the issue's 284; the difference is that I exclude stock-solution records via `record_kinds.is_solution_record`.

## Only one direction is decidable — this shapes everything

Detection rests on a finite list of component names:

- containing "yeast extract" **proves** a record is not DEFINED
- containing none of the listed names proves **nothing** — the list is not exhaustive, and an unrecognised extract simply isn't matched

So the **1,894 UNDEFINED records with no recognised undefined component are not treated as mislabelled** and are untouched. Confirming those needs per-record curation, not a word list. The audit, the repair and the corpus test are all one-directional for this reason.

## UNDEFINED vs SEMI_DEFINED is quantitative

`SEMI_DEFINED` is *"predominantly defined … supplemented with a small amount"*. Total undefined mass splits the corpus cleanly:

| undefined mass | n | action |
|---|--:|---|
| **≥ 5 g/L** | **239** | **restamped → UNDEFINED** |
| 1–5 g/L | 31 | left for curation |
| 0.5–1 g/L | 6 | left for curation |
| < 0.5 g/L | 8 | plausibly SEMI_DEFINED |
| unquantified | 1 | left for curation |

5 g/L of peptone is not "a small amount" under any reading. Whether a 1–5 g/L supplement is `SEMI_DEFINED` is a judgement about the medium, so the remaining **46** are in the report with a per-record verdict column rather than guessed at.

## The canary caught a real bug

Before touching 239 records I ran one end to end, and it **crashed**: `args.out.relative_to(REPO)` raises when `--out` points outside the repo, aborting before `--apply` did anything. Fixed with the `_display()` fallback this repo already uses elsewhere, then **re-canaried rather than fixing and fanning out in the same step**.

The second canary verified *side effects*, not the exit code: slot changed, every other line byte-identical, record still validates, report on disk and non-empty.

## Verification

- 28 new tests, including that `restamp()` leaves prose mentioning `composition_type: DEFINED` untouched
- Corpus assertion fails if any record is reverted — verified by reverting one
- `validate-strict data/normalized_yaml/bacterial` — 14,305 files, 0 errors
- Suite: 831 passed, 2 skipped

**Does not close #158** — 46 records remain for curation, and the UNDEFINED direction is deliberately unexamined.

Review follows in a comment.

### DIFF
--- a/data/import_tracking/reports/composition_type_conflicts.tsv
+++ b/data/import_tracking/reports/composition_type_conflicts.tsv
@@ -1,0 +1,47 @@
+file_path	record_id	name	undefined_components	n_undefined	undefined_g_per_l	verdict
+archaea/KOMODO_141_METHANOGENIUM_medium.yaml	CultureMech:004144	methanogenium_medium	Yeast extract; Trypticase peptone	2	3.94866	CURATOR: SEMI_DEFINED?
+archaea/KOMODO_390_PYROBACULUM_MEDIUM.yaml	CultureMech:005172	pyrobaculum_medium	Trypticase peptone; Yeast extract	2	0.69307	CURATOR: SEMI_DEFINED?
+archaea/KOMODO_927_HALORHABDUS_UTAHENSIS_medium.yaml	CultureMech:006845	halorhabdus_utahensis_medium	Yeast extract	1	1	CURATOR: SEMI_DEFINED?
+archaea/halorhabdus_utahensis_medium.yaml	CultureMech:002099	halorhabdus_utahensis_medium	Yeast extract	1	1	CURATOR: SEMI_DEFINED?
+archaea/methanococcus_sp_medium.yaml	CultureMech:004145	methanococcus_sp_medium	Yeast extract; Trypticase peptone	2	3.94866	CURATOR: SEMI_DEFINED?
+archaea/methanogenium_medium_h2_co2.yaml	CultureMech:000254	methanogenium_medium_h2_co2	Yeast extract; Trypticase peptone	2	3.94866	CURATOR: SEMI_DEFINED?
+archaea/pyrobaculum_medium.yaml	CultureMech:001497	pyrobaculum_medium	Trypticase peptone; Yeast extract	2	0.69307	CURATOR: SEMI_DEFINED?
+bacterial/NBRC_1158.yaml	CultureMech:007459	1158	Tryptone (Diffico)10gGlucose3gAgar10gDistilled water1L	1		CURATOR: unquantified
+bacterial/for_dsm_13380.yaml	CultureMech:005170	for_dsm_13380	Trypticase peptone; Yeast extract	2	0.69307	CURATOR: SEMI_DEFINED?
+bacterial/for_dsm_13514.yaml	CultureMech:005171	for_dsm_13514	Trypticase peptone; Yeast extract	2	0.69307	CURATOR: SEMI_DEFINED?
+bacterial/for_dsm_14042.yaml	CultureMech:004140	for_dsm_14042	Yeast extract; Trypticase peptone	2	3.94866	CURATOR: SEMI_DEFINED?
+bacterial/for_dsm_1497_dsm_1537_dsm_2067_dsm_2279_dsm_2373_dsm_17251_and_dsm17508.yaml	CultureMech:004129	for_dsm_1497_dsm_1537_dsm_2067_dsm_2279_dsm_2373_dsm_17251_and_dsm17508	Yeast extract; Trypticase peptone	2	3.94866	CURATOR: SEMI_DEFINED?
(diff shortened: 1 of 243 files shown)


realmarcin:
## Review — 2 findings, 1 addressed here

### 1. Restamping one slot broke agreement with the deprecated one — **fixed**

The 239 records now read `composition_type: UNDEFINED` while still carrying `medium_type: DEFINED`.

```
before   agree 11,092   disagree     0
after    agree 10,853   disagree   239
```

All the same shape: `('DEFINED', 'UNDEFINED')`.

No information was lost — `medium_type: DEFINED` was **already wrong** on these records (that is what #158 documents) and #148 mapped it across faithfully. But the corpus went from a 100% invariant to 239 violations, and "the two slots agree" is exactly the kind of thing someone would reasonably build a guard on. My change, my problem.

Realigned `medium_type: DEFINED -> COMPLEX` per the schema's own mapping. Invariant restored: **11,092 agree, 0 disagree**, plus a test.

This is the conservative choice, not the right long-term one — dropping `medium_type` entirely is probably correct now that #154 removed its last reader. Filed as **#165**; the test only holds the line until that is decided.

### 2. One ingredient name is an entire unparsed recipe — **filed as #166, not fixed**

```
Substrates*10mMKH2PO40.85gNa2HPO4·7H2O4.9g(NH4)2SO40.5gMgSO4·7H2O...
```

Found because "Tryptone" inside it matched the undefined-component scan. Corpus-wide it is the **only** such record, so it is a hand-fix rather than tooling — not done here because reconstructing its real ingredients needs the NBRC source. That record currently has no usable composition at all.

## Verified rather than asserted

| check | result |
|---|---|
| diff touches only `composition_type` lines | 239 files, **+239 / −239**, every one `DEFINED → UNDEFINED` |
| word-list precision | 6 distinct names matched, all genuine; the broad `blood`/`serum` terms matched nothing |
| corpus guard bites | reverting one record fails the test |
| `validate-strict` on `bacterial/` | 14,305 files, 0 errors |
| suite | 832 passed, 2 skipped |

The precision check mattered: `blood` and `serum` are broad enough to have caused false positives, which would have restamped correctly-DEFINED media. They matched nothing.

## Still not closing #158

46 records remain for curation (45 below the 5 g/L threshold, 1 unquantified), and the UNDEFINED direction is deliberately unexamined — absence of a recognised name proves nothing.

PR #167, opened 2026-08-03
Stop tracking the regenerated CHEBI consistency report (closes #157)

### LINKED ISSUE #157: reports/chebi_consistency.tsv is tracked but never checked, and is stale on main (#145, 4th instance)
## `reports/chebi_consistency.tsv` is tracked, regenerated by a CI gate, and stale

Running `just check-chebi-grounding` on a checkout of current `main` rewrites the
tracked report:

```
reports/chebi_consistency.tsv | 155 ++++++++--------
1 file changed, 73 insertions(+), 82 deletions(-)
```

The diff is not cosmetic — it is drift in the underlying corpus:

```
-A_name_multi_chebi  na2moo4 x 2 h2o  {'CHEBI:75213': 3196, 'CHEBI:75215': 137}
+A_name_multi_chebi  na2moo4 x 2 h2o  {'CHEBI:75213': 3224, 'CHEBI:75215': 143}
-A_name_multi_chebi  nano3            {'CHEBI:63005': 334, 'CHEBI:59148': 245}
+A_name_multi_chebi  coso4 x 7 h2o    {'CHEBI:53470': 766, 'CHEBI:91244': 5}
```

Counts have moved and whole rows have appeared and disappeared, so the committed
file describes a corpus that no longer exists. Reproduced on `main`, independent
of #148 — the axis backfill adds a `composition_type` line and cannot affect
CHEBI grounding counts.

## Why it does not fail CI

`chebi-consistency.yaml` runs `just check-chebi-grounding`, which gates on a
`--max-allowed` threshold for the *current* corpus. It never compares its output
to the committed TSV, so the file can rot indefinitely while the job stays green.
Anyone reading `reports/chebi_consistency.tsv` in the repo — rather than
re-running the gate — gets stale numbers.

## The fourth instance of #145

Same shape as the recipe indexes (#125), the id registry (#144) and the
generated dataclasses (#149): a derived artifact with no drift guard. Three of
those four are now closed; this is the remaining one.

## Fix

Either stop tracking it (it is reproducible from the corpus in one command), or
regenerate it in the gate and fail when it differs from the committed copy — the
pattern `tests/test_recipe_indexes.py` and `tests/test_id_registry.py` already
use. Tracking it *and* not checking it is the combination that produces silent
rot.

Found while reviewing #148; deliberately not fixed there, since regenerating it
would have put unrelated corpus drift into that PR's diff.

realmarcin:
## Verified the claim first

`reports/chebi_consistency.tsv` is tracked, and regenerating it on current `main` gives:

```
reports/chebi_consistency.tsv | 155 ++++++++--------
1 file changed, 73 insertions(+), 82 deletions(-)
```

Exactly the issue's numbers. The committed file describes a corpus that no longer exists, while the CI gate stays green — because it compares the *current* corpus against `--max-allowed` and never looks at the file.

## Untracked rather than guarded — this differs from the three sibling fixes

#125 (indexes) and #144 (id registry) added guards asserting the artifact matches the corpus, because those are **consumed**: things resolve ids and enumerate recipes through them, so a correct committed copy has real value.

This one has **no reader**. Nothing in `scripts/`, the justfiles, the workflows or the docs loads it — the only references are the command that *writes* it and the workflow that uploads it. And `chebi-consistency.yaml` already publishes a fresh copy on **every** run:

```yaml
- name: Upload consistency report
  if: always()
  path: reports/chebi_consistency.tsv
```

So regenerating and committing it on each corpus change would be churn on a file nobody opens. Tracking it *and* not checking it is what produced the rot; untracking removes the failure mode rather than policing it.

That is the classification #145 asked for, and it turns out to need a **third category** alongside "current view" and "dated snapshot": a *regenerable diagnostic*, which should not be an artifact at all.

## Checked the sibling

`reports/instance_validation_failures.tsv` is written the same way, so I checked it too — it is **header-only** (0 failures) and uploaded `if: failure()`. A stable empty marker rather than something that rots. Left alone.

## The guard

An untracked file cannot go stale, but it can be **re-added** by anyone who runs the recipe and then `git add -A` — which is precisely how it would come back. `tests/test_report_artifacts.py` asserts it stays untracked *and* gitignored, verified by re-adding it and watching both fail.

A third test asserts the generator still exists, since untracking is only safe while the file remains one command away. If the recipe is ever removed, the reasoning changes and the guard should be revisited rather than silently protecting nothing.

## Verification

- [x] Staleness reproduced independently on `main` (73/82)
- [x] Consumer search across scripts, justfiles, workflows, docs — none found
- [x] Guard mutation-tested by re-adding the file
- [x] `just check-chebi-grounding` still runs and writes locally
- [x] Suite: 835 passed, 2 skipped

Closes #157.

Review follows in a comment.


realmarcin:
## Review — verified the load-bearing claim, 1 issue filed

The whole PR rests on "nothing reads this file". If that is wrong, untracking is the wrong remedy and a guard was needed instead. So I checked it exhaustively rather than by the literal path:

```
grep -rniE "chebi_consistency|chebi-consistency" --include={*.py,*.sh,*.yaml,*.yml,*.md,*.json,justfile,*.justfile,*.toml}
```

Every hit is the **writer** (`project.justfile`, `audit_chebi_consistency.py`), my new test, or a reference to the *workflow name* rather than the file. **No reader.** Confirmed.

## Fresh-clone behaviour

The file no longer exists in a clone, so I removed it locally and re-ran:

```
$ just check-chebi-grounding
wrote reports/chebi_consistency.tsv (327 rows)   -> 328 lines on disk
```

Regenerates cleanly. `reports/` still exists in a clone because 115 other files are tracked there, so the directory is never missing.

## One behaviour change worth stating

The upload step is `if: always()` with no `if-no-files-found`, so it defaults to `warn`. Previously, if `check-chebi-grounding` failed *before* writing, the stale tracked copy would still have been uploaded; now the artifact is simply absent and CI warns.

I judged that an improvement rather than a regression — an absent artifact is honest, a stale one silently misleads — so I left the default rather than adding `if-no-files-found: ignore` (which its sibling `validate-strict` uses, correctly, because that file legitimately may not exist).

## Filed: #168

Auditing the siblings turned up two more tracked reports with a generator and no freshness check — `media_content_review_manifest.{tsv,json}` and `media_growth_review_manifest.{tsv,json}`. I did **not** verify whether they are stale, because doing so means running generators with side effects beyond the report. Filed rather than folded in.

That issue also records the three-category classification #145 asked for, which #157 forced into the open: consumed artifacts get guarded, dated snapshots must not be refreshed, and regenerable diagnostics with no reader should not be tracked at all.

## Verification

- [x] Staleness reproduced independently on `main` (73 insertions / 82 deletions)
- [x] Consumer search across scripts, justfiles, workflows, docs, toml — none
- [x] Fresh-clone regeneration works
- [x] Guard mutation-tested by re-adding the file — both assertions fail
- [x] Suite: 835 passed, 2 skipped

Issue #129, opened 2026-07-25
No CI workflow runs the test suite — 380 tests, 27 of them failing, never execute in CI


realmarcin:
## Problem

There are six workflows in `.github/workflows/`. **None of them runs `pytest`.**

| workflow | what it runs |
|---|---|
| `validate-strict.yaml` | `just validate-strict` |
| `chebi-consistency.yaml` | `just check-chebi-grounding` |
| `label-correspondence.yaml` | `just generate-sssom` + Engine B |
| `vendored-fleet-audit.yml` | `scripts/audit_vendored_fleet.sh` |
| `generate-pages.yaml` | page/dashboard rendering |
| `weekly-compliance.yaml` | dashboard generation |

```
$ grep -rn "pytest\|just test" .github/workflows/
(no matches)
```

A `just test` recipe exists (`project.justfile:1360`) and works — it is simply never invoked by CI. There is no other CI config (`.circleci`, `.travis.yml`, etc.) and no pre-commit pytest hook.

## Impact

**The "green checks" on a PR say nothing about whether the tests pass.** `validate-strict` and `chebi-consistency` are data-integrity gates; they do not execute `tests/`.

Consequences visible right now:

1. **27 tests are currently failing on `main`** — 16 failures + 11 errors, several in `test_kg_media_matcher.py` — and have been for some time. Nothing surfaces them.
2. Path filters make it worse: both data gates are scoped to `data/normalized_yaml/**`, `project.justfile`, and their own script. **A pure `scripts/` + `tests/` PR can land with no checks at all beyond whichever gate happens to be triggered.**
3. This is how #107 shipped a latent schema bug through a green PR — the applier wrote a `fields_changed` key that is not a slot on `CurationEvent`, which `validate-strict` would have caught only once the script had actually been *run* against the corpus. The regression test added with the fix also would not run in CI as things stand.

## Suggested fix

1. Add a `test.yaml` workflow running `uv run pytest tests/` on every PR — no path filter, or at minimum covering `scripts/**`, `tests/**`, `src/**`.
2. Triage the 27 existing failures first, or land the workflow with them explicitly xfailed/deselected and a tracking issue, so the gate starts green and stays meaningful.

Step 2 matters: wiring up pytest against a red suite produces a permanently-failing check that everyone learns to ignore, which is worse than no check.

Found while checking why #128 reported only one CI check.

PR #169, opened 2026-08-03
Gate implausible concentrations at the known backlog (#150)

realmarcin:
## Scope: one of the issue's three parts, chosen because the issue says so

#150 lists three pieces of work and names the third — a validator stopping new imports reintroducing the defect — as *"arguably the highest-leverage: without it the corpus repair is a one-off sweep that decays."* This does that piece only.

## Baselined, not zeroed

The corpus holds **11,540 implausible rows across 3,914 records**, unchanged since #135 measured them. A guard demanding zero would fail on every run and be switched off within a day — the #129 lesson about wiring a gate to a red suite.

So it baselines at the current count, the convention `check-chebi-grounding --max-allowed 101` already established in this repo: **new defects fail, the known backlog does not block**.

## Guarded in both directions

A baseline gate fails two ways, and only one is obvious:

| failure | guard |
|---|---|
| count **above** baseline — something new landed | `test_implausible_row_count_does_not_exceed_the_known_backlog` |
| baseline **far above** reality — backlog repaired, number never lowered, so the gate silently permits thousands of new defects while still passing | `test_the_baseline_is_not_far_above_reality` (fails past a 10% gap) |

Verified both bite: baseline `11_000` fails the first, `99_999` fails the second.

`--max-allowed` on the script mirrors `audit_chebi_consistency.py` so the same gate can run outside pytest. The tests reuse the existing module-scoped `corpus_findings` fixture, so this adds **no extra corpus walk**.

## Incidental: 12 stale paths in the committed reports

Refreshed both reports. The finding-set is byte-identical — 11,540 rows before and after, same records, ingredients and values — but **12 `file_path`s still said `archaea/`** for records moved to `bacterial/` by #137 and #143. Verified path-only by comparing the finding-sets keyed on (record_id, ingredient, value, finding). The #145 pattern again.

## Explicitly not done

**Items 1 and 2 of the issue.** Nesting the 579 flattened cocktails under stock `solution` objects is structural per-record surgery; the 3,335 non-cocktail records need per-value judgment. Neither is a sweep to make unilaterally on scientific data.

**#150 stays open** — this stops the decay, it does not repair the corpus.

## Verification

- [x] Gate passes at 11,540, fails at 11,539 with exit 1 and an actionable message
- [x] Both test directions mutation-tested
- [x] Report delta proven path-only, not a change in findings
- [x] Suite: 837 passed, 2 skipped

Review follows in a comment.

### DIFF
--- a/data/import_tracking/reports/concentration_plausibility.tsv
+++ b/data/import_tracking/reports/concentration_plausibility.tsv
@@ -42,6 +42,12 @@ TRACE_SALT_AS_STOCK	bacterial/8kg_4_medium.yaml	CultureMech:003180	FeCl3 x 6 H2O
 TRACE_SALT_AS_STOCK	bacterial/9k_medium.yaml	CultureMech:003277	FeSO4 x 7 H2O	49.9501	G_PER_L	49.9501 G_PER_L trace-element salt — stock-solution magnitude, final medium is normally mg/L or below
 TRACE_SALT_AS_STOCK	bacterial/CCAP_C18_BG11₀.yaml	CultureMech:000325	H3BO3	2.86	G_PER_L	2.86 G_PER_L trace-element salt — stock-solution magnitude, final medium is normally mg/L or below
 TRACE_SALT_AS_STOCK	bacterial/CCAP_C18_BG11₀.yaml	CultureMech:000325	MnCl2 x 4 H2O	1.81	G_PER_L	1.81 G_PER_L trace-element salt — stock-solution magnitude, final medium is normally mg/L or below
+TRACE_SALT_AS_STOCK	bacterial/DSMZ_1014_THIOALKALIVIBRIO_HALOPHILUS_MEDIUM.yaml	CultureMech:000243	FeSO4 x 7 H2O	2.2	G_PER_L	2.2 G_PER_L trace-element salt — stock-solution magnitude, final medium is normally mg/L or below
+TRACE_SALT_AS_STOCK	bacterial/DSMZ_1058_THIOHALOPHILUS_MEDIUM.yaml	CultureMech:000244	FeSO4 x 7 H2O	2.2	G_PER_L	2.2 G_PER_L trace-element salt — stock-solution magnitude, final medium is normally mg/L or below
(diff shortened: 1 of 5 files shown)


realmarcin:
## Review — 1 finding, addressed here

### The recipe documented a gate it did not enforce

```
# Baseline history (#150): 11,540 rows ... LOWER the baseline as records are repaired
audit-concentration-plausibility *args="":
    uv run --extra dev python scripts/audit_concentration_plausibility.py {{args}}
```

**No `--max-allowed`.** `just audit-concentration-plausibility` reported and exited 0 regardless of the count, while the comment above it read as though it gated. The convention I claimed to follow hardcodes the baseline into the recipe:

```
check-chebi-grounding *args:
    uv run python scripts/audit_chebi_consistency.py --out ... --max-allowed 101 {{args}}
```

The gate *did* exist — in the pytest test, which runs on every PR with no path filter — so the protection was real and only the documentation was wrong. But there was no non-pytest path enforcing it, and the recipe is in no workflow.

### The fix creates a hazard, so it is guarded too

Wiring `--max-allowed 11540` into the recipe puts the number in **two** places that must be lowered together as the backlog is repaired. Miss one and either the gate stops biting or local runs fail while CI passes — the same drift class as #144 and #157.

So `test_justfile_baseline_matches_the_test_baseline` parses the baseline out of `project.justfile` and asserts it equals `CONCENTRATION_BACKLOG_BASELINE`. Verified it fails on **both** ways this can rot:

| mutation | result |
|---|---|
| recipe gates at a different number | FAILED |
| recipe drops `--max-allowed` entirely | FAILED |

Filed as **#170** rather than fixed silently — "two hardcoded copies of a number that is expected to change" is a pattern this repo has now produced four times, and the issue is the record that it was a deliberate trade rather than an oversight.

### Verified

| check | result |
|---|---|
| gate passes at 11,540, fails at 11,539 | exit 1, actionable message |
| both baseline directions bite | `11_000` fails count-check; `99_999` fails staleness-check |
| justfile/test drift guard bites | both mutations fail |
| report delta is path-only | finding-set byte-identical on (record_id, ingredient, value, finding); 12 `archaea/` → `bacterial/` paths from #137/#143 |
| suite | 838 passed, 2 skipped |

### Unchanged

**#150 stays open.** Items 1 and 2 — nesting 579 flattened cocktails under stock `solution` objects, and 3,335 non-cocktail records needing per-value judgment — are structural surgery and scientific judgement respectively. This stops the decay; it does not repair the corpus.

PR #171, opened 2026-08-03
Populate SEMI_DEFINED on evidence, not assumption (#152)

### LINKED ISSUE #152: nutritional_class and functional_role ship populated on 1 record each; SEMI_DEFINED on none
## What ships in #148

#148 replaces the single-valued `medium_type` with three orthogonal axes. Only
one of them actually has data.

Post-merge counts across the 15,878-record corpus:

| axis | records populated |
|---|---|
| `composition_type` | 11,092 |
| `nutritional_class` | **1** |
| `functional_role` | **1** |

The one record in each case is `data/normalized_yaml/bacterial/r2a_medium.yaml`,
set by hand in the R2A curation commits — not by the migration.

## Why

The migration can only derive axes from `medium_type`, and the corpus never used
the values that feed the other two axes:

```
8612 COMPLEX
2480 DEFINED
   1 NEGATIVE_CONTROL
   1 BUFFER
```

`MINIMAL`, `SELECTIVE`, `DIFFERENTIAL` and `ENRICHMENT` are permissible values of
`MediumTypeEnum` but appear on **zero** records. So the migration's mappings for
them — MINIMAL→`nutritional_class`, and the three functional values→
`functional_role` — never fired even once. This is not a bug in the migration;
there was simply nothing to convert.

## Why it matters

The schema now advertises three axes and a KGX consumer can filter on any of
them, but two return an empty set. `SEMI_DEFINED` is likewise a permissible
`composition_type` value that nothing emits — every undefined-component medium
is flatly `UNDEFINED`, including media that are a defined base plus a trace of
yeast extract.

The gap is worth stating explicitly because the corpus *looks* fully migrated:
11,092 of 11,094 typed records carry an axis value, so a coverage check on
`composition_type` alone reads as complete.

## What would close it

These need signal the old slot never had, so they are curation work, not another
mechanical pass:

- `nutritional_class` — inferable in bulk for well-known media (M9 and the
  defined-minimal family → MINIMAL; LB/TSB/BHI → RICH). Most of the rest are
  plausibly GENERAL_PURPOSE but that should not be assumed by default.
- `functional_role` — needs the recipe or the source catalogue (presence of
  antibiotics, dyes, bile salts, or a differential indicator).
- `SEMI_DEFINED` — separable from UNDEFINED by ingredient analysis: a
  predominantly defined ingredient list with exactly one small undefined
  component.

Filed from the review of #148. Not a blocker for that PR — the schema and the
composition backfill are correct as far as they go.

realmarcin:
## Verified before starting

#152 reports three composition axes shipped in #148 and only one carrying data. Confirmed on `main`:

| axis | records |
|---|--:|
| `composition_type` | 11,092 |
| `nutritional_class` | **1** |
| `functional_role` | **1** |
| `composition_type: SEMI_DEFINED` | **0** |

This addresses the SEMI_DEFINED third. **612 records promoted**; `composition_type` is now DEFINED 2,241 / UNDEFINED 8,239 / **SEMI_DEFINED 612**.

## The hard half is proving "predominantly defined"

`SEMI_DEFINED` needs *both* a small undefined amount **and** a predominantly defined base. A mass threshold gets the first half only.

**Presence of an undefined component is provable; absence is not** — the name list is finite. Treating "nothing else matched" as proof would run the #158 asymmetry backwards and silently promote a record carrying an unrecognised extract at 10 g/L.

So promotion requires *positive* evidence: **every other ingredient must carry a CHEBI id.**

```
809  UNDEFINED, exactly one undefined component < 0.5 g/L
612  ...and every other ingredient CHEBI-grounded   -> promoted
197  ...'predominantly defined' merely assumed      -> left alone
```

The promoted set is the enum's own example — mineral-salts media with a trace of yeast extract:

```
JCM_J176_ACIDIANUS_BRIERLEYI_MEDIUM      7 ingredients | Yeast extract 0.2  g/L
KOMODO_1091_HALOQUADRATUM_WALSBYI        12 ingredients | Yeast extract 0.05 g/L
JCM_J283_FERROPLASMA_ACIDIPHILUM         6 ingredients | Yeast extract 0.16 g/L
```

Canaried one record end to end before the batch: slot changed, every other line byte-identical, record still validates.

## My own guard caught this change

The #164 guard failed on the promotion: the 612 now read `medium_type: COMPLEX` / `composition_type: SEMI_DEFINED`, which its one-to-one mapping called a contradiction.

It isn't. The deprecated `MediumTypeEnum` has **no SEMI_DEFINED** — it is coarser — and `COMPLEX` meant "contains an undefined component", which these do. So `COMPLEX` legitimately maps to `{UNDEFINED, SEMI_DEFINED}`; the mapping was too strict once a third value became populated. `DEFINED` stays one-to-one, since it still means "no undefined component".

Verified the widened mapping **still fails on a genuine contradiction** rather than just accepting everything.

## Does not close #152

`nutritional_class` and `functional_role` remain at 1 record each. Both need signal the corpus does not carry — catalogue provenance, or antibiotic/dye/indicator detection — and neither is derivable the way this was.

## Verification

- [x] All three of the issue's counts reproduced independently
- [x] 6 new detector tests, including that an ungrounded sibling ingredient blocks promotion
- [x] Corpus test asserting all three `composition_type` values are non-empty, so SEMI_DEFINED cannot silently regress to zero
- [x] `validate-strict` archaea/ — 743 files, 0 errors
- [x] Suite: 845 passed, 2 skipped

Review follows in a comment.


realmarcin:
## Review — 1 finding, which exposed a second; both addressed

### The promotion left no record of what it declined — **fixed, filed as #172**

612 records promoted, **197 held back** solely because a sibling ingredient is ungrounded. Those 197 are the actionable subset, and they were recorded **nowhere** — aggregate counts to stdout, then exit. `composition_type_conflicts.tsv` covers only the opposite direction.

They matter because the reason is decidable per record, and the first rows show both outcomes:

```
Sodium phosphate buffer    -> defined chemical that needs grounding
Cystein-HCl x 2 H2O        -> defined chemical that needs grounding
Polar lipid fraction       -> genuinely undefined; the word list should learn it
```

So a near-miss is either a grounding task or a word-list gap, and the report now names the sibling so a curator can tell which.

### Writing that report exposed a defect I had shipped

Scanning only `UNDEFINED` records meant a **re-run dropped every PROMOTED row** — those records are `SEMI_DEFINED` by then. The committed artifact would have shown 197 near-misses and no trace of the 612 promotions, changing shape between the promoting run and the next.

Now scans both states. `restamp()` is a no-op when the value already matches, so the promoted **count** stays honest at 0 on a re-run while the **rows** persist. Verified byte-identical across consecutive runs: **612 PROMOTED + 197 CURATOR**.

## Verified rather than asserted

| check | result |
|---|---|
| corpus diff surgical | 612 files, **+612 / −612**, every one `UNDEFINED → SEMI_DEFINED` |
| widened mapping not weakened | still **rejects** `DEFINED`+`SEMI_DEFINED`, `DEFINED`+`UNDEFINED`, `COMPLEX`+`DEFINED`; accepts only the two legitimate `COMPLEX` pairings |
| no overlap with #164's 239 records | 0 — as predicted, those carry ≥5 g/L and these <0.5 g/L |
| report idempotent | byte-identical across runs |
| suite | 846 passed, 2 skipped |

The mapping check was the one worth doing: widening a guard to accommodate your own change is exactly how a guard quietly stops guarding. It still rejects 4 of 6 combinations.

## Unchanged

**#152 stays open.** `nutritional_class` and `functional_role` remain at 1 record each — both need signal the corpus does not carry, and neither is derivable the way SEMI_DEFINED was.

PR #173, opened 2026-08-03
Rank resolver matches by specificity instead of raising on all of them (#151 item 3)

### LINKED ISSUE #151: Per-pair curation of the 290 cross-category filename collisions (unfinished half of #116)
Splitting the unfinished half of #116 into its own issue.

#116 asked for per-pair reconciliation of media sharing a filename across category directories. **#134 shipped a deliberately read-only triage** and says so explicitly ("This does not close #116"). #137 then did the bacterial-half curation and commented to the same effect. #116 was closed on 2026-07-26 anyway. This issue tracks the curation that remains.

## What is already resolved

- **The 11 named `methano*` collisions** — #120 moved all 11 to `archaea/` only, pinned by a test so they cannot come back.
- **The bacterial half** — #137 moved 27 recipes from `archaea/` to `bacterial/` (*Thermus*, *Pseudomonas*, *Treponema*, *Spirochaeta*, …), taking `archaea/` from 773 to 746.

## What #134 actually found — bigger than #116 describes

**290 filenames appear in two or more category directories**, not the ~50 the issue implies. Re-counted from the committed `data/import_tracking/reports/filename_collisions.tsv` on 2026-07-30:

| classification | pairs | meaning |
|---|--:|---|
| `DIFFERENT` | 223 | compositions genuinely diverge |
| `IDENTICAL` | 56 | verbatim same labels, groundings, concentrations, units |
| `EQUIVALENT` | 11 | same medium, differing only in ingestion artefacts |

## Two things to know before picking this up

**Even the 56 IDENTICAL pairs are not safe to auto-dedupe.** Choosing which copy survives means choosing which category is correct — a domain judgment, the same one #115/#120/#137 kept getting wrong in bulk — and deduping means deleting a CultureMech id. This needs per-pair curation, not a script.

**118 of the 223 DIFFERENT pairs share no ingredient rows at all.** Those are unrelated media that merely collide on filename; they need disambiguation (rename one side), not reconciliation. Only 3 pairs overlap at 90% or more. So the 223 splits into a large trivial-rename group and a small genuinely-hard group, and separating them is cheap.

## Why the EQUIVALENT tier exists

A tuple comparison is too blunt. `1_10_sabourauds_agar.yaml` has identical concentrations throughout, yet `bacterial/` grounds Glucose to `CHEBI:42758` while `fungal/` grounds it to `CHEBI:17234`; labels differ (`MgSO4・7H2O` vs `MgSO4 x 7 H2O`); and the `bacterial/` copy carries an extra implausible `Distilled water 1 G_PER_L` row (see the #118 follow-up). Reconciling these means resolving a grounding disagreement, not just picking a file.

## Also worth carrying forward

#137's reverse audit of `archaea/` found the original 71-file import had never been domain-vetted at all — 29 genuinely archaeal, 26 genuinely bacterial, 16 ambiguous. That argues for keeping a permanent domain check rather than sweeping once, which is the same conclusion #145 reaches from a different direction.

## Remaining work

1. Split the 223 `DIFFERENT` pairs into the 118 no-overlap renames and the rest.
2. Curate the 56 `IDENTICAL` and 11 `EQUIVALENT` pairs per pair, deciding category and surviving id.
3. Disambiguate the resolver's multi-match behaviour (it hit `syntrophomonas_medium_for_syntrophospora_cellicola_19j_3`, which has a `TOGO_M520_*` sibling of the same recipe).

realmarcin:
## The bug is much larger than its one reported symptom

#151 records the resolver raising on `syntrophomonas_medium_for_syntrophospora_cellicola_19j_3`. Measuring it:

```
2,291 record `name:` values are shared by more than one record
...and every one of them is also some other record's filename
```

`resolve_media_file` treated every identifier match as equally good, so `lb_broth` matched `bacterial/lb_broth.yaml` **by filename** and `bacterial/TOGO_M3227_LB_broth.yaml` **by its `name:` field**, and raised. **Any slug-driven batch failed on all 2,291.**

Found concretely rather than by audit: it blocked the first dry-run of the #152 axis research, on its very first target.

## Fix: rank, don't refuse

Matches group into tiers; the most specific tier that matches wins:

| tier | uniqueness |
|---|---|
| 1. CultureMech id | unique by construction, guarded by `tests/test_id_registry` |
| 2. filename stem / full filename | unique within a directory |
| 3. `name` / `original_name` / `media_term.preferred_term` | **not** unique |

Sampling 400 previously-ambiguous slugs: **374 now resolve, 26 still raise.**

## The 26 that still raise are the point

Every one is a genuine cross-directory filename collision from #151:

```
potato_dextrose_agar  -> bacterial/ + fungal/
bg11                  -> bacterial/ + algae/
jm                    -> bacterial/ + algae/
s88_vitamins          -> bacterial/ + algae/
```

No tier can separate two files with the same stem, so raising is correct. Silently picking one would hide exactly the collisions #151 exists to track — so a test pins that behaviour against a future "improvement" that resolves them.

The id tier is the escape hatch, tested on **both sides** of a collision (`CultureMech:009646` → `lb_broth.yaml`, `CultureMech:009666` → `TOGO_M3227_LB_broth.yaml`) so neither record becomes unreachable.

`_candidate_paths` is kept, now flattening the tiers, since six scripts import from this module.

## Verification

- [x] The exact Edison dry-run that failed now succeeds
- [x] 5 new tests: filename-beats-field, id reaches both sides, the #151 case, genuine collision still raises, absent target still `FileNotFoundError`
- [x] Suite: 851 passed, 2 skipped

**Does not close #151** — items 1 and 2, the per-pair curation of the 274 collisions, remain. This is item 3 only.

Review follows in a comment.

### DIFF
--- a/scripts/research_media.py
+++ b/scripts/research_media.py
@@ -33,40 +33,82 @@ def _normal_key(value: str) -> str:
 
 
 def _candidate_paths(target: str) -> list[Path]:
+    """All records matching `target` on any identifier. Kept for callers that want
+    the full set; `resolve_media_file` ranks these rather than treating them as
+    equally good."""
+    return [p for tier in _tiered_candidates(target) for p in tier]
+
+
+def _tiered_candidates(target: str) -> list[list[Path]]:
+    """Matches grouped by how specific the match is, most specific first.
+
+    Ranking matters because 2,291 record `name:` values are shared by more than one
+    record, and every one of them is also some other record's filename. So
+    `lb_broth` matches both `bacterial/lb_broth.yaml` (by filename) and
+    `bacterial/TOGO_M3227_LB_broth.yaml` (by its `name:` field). Treating those as
+    equally good made the resolver raise on 2,291 slugs, which blocked any
+    slug-driven batch (#151).
+
+    Tiers, in order:
+      1. CultureMech id — unique by construction, guarded by tests/test_id_registry.
+      2. filename stem or full filename — unique within a directory, and the
+         corpus keeps filenames unique across directories except for the
+         collisions tracked in #151.
+      3. `name` / `original_name` / `media_term.preferred_term` — NOT unique.
+    """
     target_path = Path(target)
     if target_path.exists():
-        return [target_path.resolve()]
+        return [[target_path.resolve()]]
 
     normalized_target = _normal_key(target)
-    matches = []
+    by_id: list[Path] = []
+    by_filename: list[Path] = []
+    by_field: list[Path] = []
+
     for path in sorted(MEDIA_DIR.glob("*/*.yaml")):
-        if _normal_key(path.stem) == normalized_target or _normal_key(path.name) == normalized_target:
-            matches.append(path)
+        if _normal_key(path.stem) == normalized_target or \
+                _normal_key(path.name) == normalized_target:
+            by_filename.append(path)
             continue
+        # Cheap reject before parsing: the identifier must appear somewhere.
         if normalized_target not in path.read_text().casefold():
             continue
         doc = load_media(path)
+        if doc.get("id") is not None and \
+                _normal_key(str(doc["id"])) == normalized_target:
+            by_id.append(path)
+            continue
         media_term = doc.get("media_term")
         identifiers = [
-            doc.get("id"),
             doc.get("name"),
             doc.get("original_name"),
             media_term.get("preferred_term") if isinstance(media_term, dict) else None,
         ]
-        if any(value is not None and _normal_key(str(value)) == normalized_target for value in identifiers):
-            matches.append(path)
-    return matches
+        if any(v is not None and _normal_key(str(v)) == normalized_target
+               for v in identifiers):
+            by_field.append(path)
+
+    return [tier for tier in (by_id, by_filename, by_field) if tier]
 
 
 def resolve_media_file(target: str) -> Path:
-    """Resolve a path, slug, CultureMech ID, name, or original name to one media YAML file."""
-    matches = _candidate_paths(target)
-    if len(matches) == 1:
-        return matches[0]
-    if len(matches) > 1:
-        choices = ", ".join(str(path.relative_to(REPO_ROOT)) for path in matches[:20])
-        raise ValueError(f"Target matched multiple media records: {choices}")
-    raise FileNotFoundError(f"Media target not found under {MEDIA_DIR}: {target}")
+    """Resolve a path, slug, CultureMech ID, name, or original name to one media YAML.
+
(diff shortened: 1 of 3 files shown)


realmarcin:
## Review — 1 issue filed, no changes needed to the fix

### Safety: does ranking silently change what existing callers get?

This is the risk that matters — six scripts import this resolver, and a change that quietly returns a *different* record would be worse than the bug.

It cannot, and the argument is structural rather than statistical:

- **Old:** returned a path only when the flat match list had exactly one element.
- **New:** `_candidate_paths` is now the flattened tiers, so if the flat list had exactly one element, that element sits alone in exactly one tier — which is therefore `tiers[0]`, and the new code returns the same path.

So the change is **monotone**: identical answers wherever it previously worked, and failures converted to successes. `FileNotFoundError` behaviour is unchanged (empty tiers).

Confirmed empirically on 50 sampled targets (filenames and CultureMech ids):

```
same=45   changed=0   lost=0   gained=5
```

A note on that check: my first attempt reported `gained=50, same=0`, which would have meant the old resolver failed on everything. It was my harness — loading the old module from `/tmp` made its `REPO_ROOT = Path(__file__).parent.parent` resolve to `/`, so it saw no corpus. Repointing `MEDIA_DIR` gave the numbers above. Worth recording because I nearly filed that as a finding.

### Filed: #174 — the resolver is O(corpus) per call

```
resolve("lb_broth")                 0.99s
resolve("CultureMech:009646")       0.73s
resolve("data/.../lb_broth.yaml")   0.00s   (path short-circuit)
```

Every non-path resolution re-reads ~15,900 files. This PR neither causes nor worsens it — the old resolver was identical — but it becomes newly relevant, because the whole point here is to make slug-driven **batches** work, and a 500-target batch now spends ~8 minutes resolving before its first API call.

Not bundled in: the fix is orthogonal to the correctness change, and the tiered structure introduced here makes an index straightforward later (the tiers are just three dicts).

It also made this very review awkward — the 600-target comparison exceeded a 10-minute timeout, which is why the sample above is 50.

### Verified

| check | result |
|---|---|
| monotonicity | proved structurally, confirmed on 50 targets |
| the #151 case resolves | `syntrophomonas_…_19j_3.yaml` |
| id reaches both sides of a collision | `009646` → `lb_broth.yaml`, `009666` → `TOGO_M3227_LB_broth.yaml` |
| genuine collisions still raise | 26/26 are cross-directory filename clashes |
| the blocked Edison dry-run | now succeeds |
| suite | 851 passed, 2 skipped |

**Does not close #151** — items 1 and 2, per-pair curation of the 274 collisions, remain.

Issue #175, opened 2026-08-03
226 media have no usable composition (was 463, then 428 — 202 were mis-typed stock solutions)


realmarcin:
## Surfaced by the review-need ranking

Two adjacent failure modes, both meaning the record carries no recipe at all:

```
362  records with zero ingredients
101  records whose only ingredient text is "See source for composition"
---
463  records (4.2% of 11,094 media) with no usable composition
```

The 101 are strikingly uniform — every one uses that exact string, so this is one import path, not scattered curation gaps.

## Why this was not visible before

`prioritize_deep_research_candidates.py` **hard-filters zero-ingredient records** before scoring, and rewards recipe completeness thereafter. That is correct for its purpose — ranking where deep research will pay off — but it means the damaged tail is invisible to the only ranking the repo had. These 463 records have never appeared at the top of anything.

`validate-strict` does not catch them either: `ingredients` is not required, so an empty list is schema-valid.

## Included in the 463

`bacterial/test_medium_123.yaml` — `original_name: "Test Medium 123"`, sole ingredient "See source for composition", holding `CultureMech:000306`. A test fixture that reached the production corpus and was assigned a real id.

## What to decide

Per group, and they may differ:

1. **Recoverable?** The `notes` on several say "Imported from BacDive cultivation data", so the source may still have the composition and a re-import could fill them.
2. **Or honestly empty?** If the source genuinely has no composition, the record should say so — a `data_quality_flags` entry — rather than carrying one fake ingredient that pollutes every text scan. The placeholder string already matches ingredient searches, which is how it turned up here.
3. **Or spurious?** `test_medium_123` should probably not exist. Deleting it means retiring a CultureMech id, which is a separate decision (cf. #144).

## Reproduce

```
just score-review-need --top 40
```

They occupy the 70-80 band. Ranking, signals and weights are in `scripts/score_review_need.py`.

realmarcin:
**Re-measured in #193 (merged): 428, not 463.**

35 records carry their whole composition under `solutions`; the original count
read `ingredients` alone. Same slot that made 15 of 17 findings in #181 false
positives.

`just triage-missing-compositions` now classifies them:

```
327  KOMODO ModelSEED, nothing in either slot
101  a single placeholder ingredient
```

**202 of the 428 are named like a stock solution, not a medium** — "Trace element
solution (medium 929)", "Haemin solution (medium 104)", "10 x M9 salts",
[...]

Issue #166, opened 2026-08-02
One ingredient name is an entire unparsed recipe: 'Substrates*10mMKH2PO40.85gNa2HPO4·7H2O4.9g...'


realmarcin:
## A whole recipe crammed into a single `preferred_term`

`data/normalized_yaml/.../NBRC_1197.yaml` has an ingredient whose name is:

```
Substrates*10mMKH2PO40.85gNa2HPO4·7H2O4.9g(NH4)2SO40.5gMgSO4·7H2O...
```

That is the source recipe's whole composition block — several ingredients, each with its own quantity and unit — concatenated into one field with the whitespace stripped. It is an import parse failure, not a real ingredient.

## Scope: this appears to be the only one

Scanning every non-solution record for a `preferred_term` longer than 40 characters containing two or more embedded quantity+unit pairs:

```python
MANGLED = re.compile(r"\d+\s*(g|mg|ml|l)\b.*\d+\s*(g|mg|ml|l)\b", re.I)
# -> 1 hit, corpus-wide
```

So this is a single bad record rather than a systemic pattern — cheap to fix by hand, and worth fixing precisely because it is isolated enough not to need tooling.

## Why it matters beyond one record

1. The record's real ingredients are **not represented** — they exist only inside this string, so the medium has no usable composition.
2. It pollutes text matching. It was found because the substring "Tryptone" inside it matched an undefined-component scan in #164; anything else grepping `preferred_term` will match it spuriously too.
3. It will not be caught by concentration or grounding audits, because the field it corrupts is the name.

## Suggested fix

Split it back into proper ingredient entries from the NBRC source, or — if the source is not readily available — replace it with a single honest placeholder and flag the record as having an unparsed composition, so downstream consumers see "unknown" rather than one nonsense ingredient.

A cheap guard would be a test asserting no `preferred_term` contains two or more embedded quantity+unit pairs. It would have caught this at import.

Found while auditing composition_type in #164.

realmarcin:
## Correction: 25 records, not 1 — and it is a clean NBRC cluster

This issue says the scan found "1 hit, corpus-wide" and concludes it is "a single bad record rather than a systemic pattern", so a hand-fix. **Both of those are wrong**, and the fault is in the regex I used when filing.

The scan used:

```python
MANGLED = re.compile(r"\d+\s*(g|mg|ml|l)\b.*\d+\s*(g|mg|ml|l)\b", re.I)
```

[...]

PR #176, opened 2026-08-03
Rank media by need for review, the inverse of the research-yield ranking

realmarcin:
## The design point

`prioritize_deep_research_candidates.py` ranks by research **yield** — it rewards completeness and source recognizability, and **hard-filters zero-ingredient records**. That is right for its purpose, but it means the records it discards are exactly the ones needing review. Running only that ranking is how **362 zero-ingredient records sat unexamined**.

This scorer asks the opposite question: which records look broken?

## Signals were measured before weights were chosen

So a signal firing on half the corpus cannot outweigh one firing on 3%:

| signal | weight | records |
|---|--:|--:|
| no ingredients | 30 | 362 (3.3%) |
| placeholder ingredient text | 25 | 167 (1.5%) |
| unparsed recipe in a name | 25 | 25 (0.2%) |
| only 1–2 ingredients | 15 | 627 (5.7%) |
| no ingredient grounded | 20 | 398 (3.6%) |
| under half grounded | 10 | 537 (4.8%) |
| no `media_term` | 10 | 695 (6.3%) |
| strain-pointer name | 10 | 248 (2.2%) |
| no notes | 5 | 150 (1.4%) |
| no pH and no temperature | 5 | 5,704 (51.4%) — deliberately low |

**42 records score 70+**, 151 score 50–69. The top is independently confirmed broken: `NBRC_1197` (an entire recipe in one ingredient name) and `test_medium_123` — a test fixture holding `CultureMech:000306` in the production corpus.

## Two features measured, found useless, and dropped

- **"no conditions block"** fired on **100%** — the slots are `ph_value`/`temperature_value`, not a block.
- **first-pass "uninformative name"** fired on **72%**, flagging real media like `1/2 Erdschreiber Medium`.

Short names are not a signal either: **BG11** and **JM** are real media, and a draft flagging them caught 284 records, mostly legitimate. A test pins that.

## The tests hardened a regex that would have shipped weak

The mangled-name pattern used `\b` after the unit, but these names are concatenated with separators stripped — `0.85gNa2HPO4` has no word boundary after the `g`. It matched `NBRC_1197` **only by luck**, because a `(` happens to follow one unit further along its 272-character name.

Corrected, it finds **25 records, every one NBRC**. That corrects #166, which I filed claiming "1 hit, corpus-wide… a single bad record rather than a systemic pattern" — commented there.

## Deliberately not scored

Implausible concentrations, composition_type contradictions and filename collisions each already have a dedicated audit. Duplicating that logic here would let the two drift, so this scorer has **one owner per signal** and no flag to merge them; join on `file_path` instead.

## Also filed

**#175** — 463 records (4.2%) with no usable composition: 362 empty, 101 whose only ingredient is the literal string "See source for composition".

## Verification

- [x] 15 tests, including that a healthy record scores **0** — a ranking that flags everything ranks nothing
- [x] Corpus tests: known-bad records rank in the worst 60; fewer than 500 score ≥50
- [x] Regex checked against both false-positive and false-negative cases
- [x] Suite: 866 passed, 2 skipped

Review follows in a comment.


realmarcin:
## Review — 1 finding, addressed here

### The report was 60% of the corpus — **fixed, filed as #177**

```
band     n      dominant signals
70+      66     no ingredient grounded x66, no media_term x66
50-69   127     no ingredient grounded x127, only 1 ingredient x93
30-49   632     no pH and no temperature x460, no ingredients x362
15-29   940     no pH and no temperature x805, no media_term x348
1-14   4878     no pH and no temperature x4332
```

**4,332 of the 4,878 low-band records (89%) carried *only* "no pH and no temperature"** — a signal firing on 51.4% of media.

I had already weighted it lowest at 5 *because* it is the norm. That was not enough: any non-zero score emitted a row, so a record with nothing wrong except a missing optional slot sat in a file titled "needs review".

That reproduced, one layer up, the exact failure this scorer exists to fix. The yield ranking hides the damaged tail by filtering it out; a 60%-noise report hides it by burying it — the 42 genuinely broken records were under 4,332 that merely lack a pH value.

Records now qualify only on a signal outside `NORM_LEVEL_SIGNALS`; conditions still contribute to the score, so they break ties among already-suspect records.

```
6,643 -> 2,311 rows
conditions-only rows: 4,332 -> 0
records scoring >=50:   193 -> 193      (nothing substantive dropped)
```

The low band is now substantive as well — its commonest reasons are "no media_term" and partial grounding.

### Verified

| check | result |
|---|---|
| nothing substantive dropped | ≥50 band unchanged at 193 |
| both halves of the rule pinned | conditions-only not emitted; conditions still break ties |
| known-bad records still rank top | `NBRC_1197`, `test_medium_123` in the worst 60 |
| healthy record scores 0 | yes — a ranking that flags everything ranks nothing |
| suite | 868 passed, 2 skipped |

My first version of the conditions-only test **failed**: the fixture had one ingredient, which tripped the "only 1–2 ingredients" signal, so it was not conditions-only. The fixture was wrong, not the code — corrected to three.

Issue #118, opened 2026-07-22
Ingredient mis-normalization: trace-element stock-solution values and unit errors stored as final medium concentrations


realmarcin:
## Summary

Deep-research review of several media surfaced a recurring data-quality pattern: **trace-element stock-solution concentrations and mis-scaled units are stored as final per-liter medium concentrations.** These are not research gaps — they are normalization bugs in the recipe corpus that will corrupt any downstream concentration analysis or KG export.

### 1. `archaea/sulfolobus_medium_for_dsm_9790.yaml` (CultureMech:008911)
Trace-element and water entries are dimensionally impossible as final-medium values:

| Ingredient | Stored value | Problem |
|---|---|---|
| Distilled water | `2000 G_PER_L` | Looks like a **2-L preparation volume**, not a concentration |
| MnCl2·4H2O | `180 G_PER_L` | Trace-element **stock** value, not final medium |
| Na2B4O7·10H2O | `450 G_PER_L` | Trace-element stock value |
| ZnSO4·7H2O | `22 G_PER_L` | Trace-element stock value |
| CuCl2·2H2O | `5 G_PER_L` | Trace-element stock value |
| Na2MoO4·2H2O | `3 G_PER_L` | Trace-element stock value |

Published modified DSMZ Medium 88 formulations (e.g. Leung et al. 2024, doi:10.1038/s41467-024-47324-2; Auernik et al. 2008, doi:10.1128/aem.02019-07) are far simpler; these high-concentration entries clearly belong in a **trace-element stock/solution object with an addition volume**, not as direct g/L ingredients.

### 2. `bacterial/TOGO_M1791_Pelobacter_acetylenicus_Medium.yaml` (CultureMech:008359)
Same pattern — implausibly high trace-salt g/L values that read as stock-solution concentrations rather than final medium.

### 3. `bacterial/TOGO_M1796_Desulfovibrio_medium.yaml` (CultureMech:008364) — unit error
Resazurin (a redox indicator) is stored as:
```yaml
- preferred_term: Resazurin
  concentration:
    value: '1'
    unit: G_PER_L      # <-- standard use is ~1 mg/L; this is 1000x too high
```

## Why it matters

[...]

PR #179, opened 2026-08-04
Report the 25 unparsed NBRC compositions — and deliberately do not auto-repair them (#166)

### LINKED ISSUE #166: One ingredient name is an entire unparsed recipe: 'Substrates*10mMKH2PO40.85gNa2HPO4·7H2O4.9g...'

realmarcin:
## What these records are

25 `NBRC_*` records carry their entire composition crammed into the first ingredient's `preferred_term`, separators stripped, plus an empty second ingredient:

```
Tryptone5gYeast extract3gNaCl10gUrea1gDistilled water1LAgar (if needed)15g
```

## I tried to repair them mechanically, and the attempt is why this reports instead

Two checks looked sufficient:

- **Round trip** — reassembling `name + value + unit` for every item, plus the tail, must reproduce the source **exactly**. This proves the parse invented nothing, and it caught `Tween 800.3g` being read as `Tween` + `800.3`.
- **Plausibility** — no non-water solid above 100 g/L.

**20 of 25 passed both, were applied, and validated cleanly.** Then I read the output:

```
KH2PO   at 40.85 G_PER_L      <- what I wrote
KH2PO4  at  0.85 G_PER_L      <- the truth
```

`KH2PO4` + `0.85g` concatenates to `KH2PO40.85g`, which splits **equally well** into `KH2PO` + `40.85g`. Both round-trip. 40.85 is under the plausibility ceiling. The parse had written a **non-existent compound at 47× the real concentration**, in **11 of the 20 records** — across KH2PO4, K2HPO4, CaCO3 and H3BO3.

All 20 edits were reverted. **The corpus in this PR is byte-identical to main** — verified.

## Why a third check cannot rescue it

A formula-shaped name ending in an uppercase letter now flags the hazard. But it can only **refuse**, never correct, because it matches `MgO` just as readily — and MgO is a real compound.

That asymmetry is the argument. `CaCO32mg` is not resolvable from the string: it is `CaCO3` + `2mg` or `CaCO` + `32mg`, and only chemistry decides. That needs the NBRC catalogue, not a heuristic.

So the tool reports and **cannot write** — a test asserts no write path exists.

## What it delivers

25 opaque records become a structured worklist: proposed parse, recovered trailing note, and the specific reason each held record is ambiguous. **14 PROPOSED, 11 HOLD.**

## One more defect the attempt surfaced

An early version wrote `Distilled water1L` as `1 ML_PER_L` — inventing the exact defect #118 documents, water recorded as a solute, in a unit the concentration gate does not even check. A volume basis is now carried in notes instead.

`Seawater750ml` is different and legitimate — a proportion, distinguished by the **unit**, not by the word "water".

## Verification

- [x] Corpus byte-identical to main
- [x] 14 tests, including that the tool has no write path and that the truncated-formula rule only refuses
- [x] Every one of the 25 records gets a verdict with a recorded reason
- [x] Suite: 865 passed, 2 skipped

**Does not close #166** — the records still need a curator with the source catalogue.

Review follows in a comment.

### DIFF
--- a/data/import_tracking/reports/unparsed_compositions.tsv
+++ b/data/import_tracking/reports/unparsed_compositions.tsv
@@ -1,0 +1,26 @@
+file_path	record_id	verdict	n_recovered	detail	trailing_note
+bacterial/NBRC_1063.yaml	CultureMech:007453	HOLD: parse does not round-trip	0	reassembly differs from the source; the split is ambiguous (e.g. 'Tween 800.3g' — is 80 a grade or part of the value?)	Tris-HCl buffer.
+bacterial/NBRC_1135.yaml	CultureMech:007458	PROPOSED	4	4 ingredients parsed	*Manufactured by Asahi Breweries, Ltd., Tokyo, Japan. Distributed by Tanabe Seiyaku Co. Ltd., Osaka, Japan.
+bacterial/NBRC_1158.yaml	CultureMech:007459	PROPOSED	4	4 ingredients parsed	
+bacterial/NBRC_116.yaml	CultureMech:007460	PROPOSED	8	8 ingredients parsed	
+bacterial/NBRC_1197.yaml	CultureMech:007463	HOLD: possible truncated formula	15	KH2PO, Na2HPO4·7H2O, MgSO4·7H2O, FeSO4·7H2O, MgO, CaCO, ZnSO4·7H2O, MnSO4·4H2O, CuSO4·5H2O, CoSO4·7H2O, H3BO — a trailing subscript may have been absorbed into the value; both splits round-trip	*Vanillic acid and syringic acid etc.
+bacterial/NBRC_1210.yaml	CultureMech:007464	HOLD: possible truncated formula	9	K2HPO, MgSO4·7H2O — a trailing subscript may have been absorbed into the value; both splits round-trip	Distilled water1LpH 7.2+/-0.2
+bacterial/NBRC_1287.yaml	CultureMech:007469	PROPOSED	4	4 ingredients parsed	Agar15gpH 7.0Biphenyl crystal is supplied on a lid of petri dish as vapor phase.
+bacterial/NBRC_1379.yaml	CultureMech:007475	HOLD: implausible value	5	Sucrose 500g	
+bacterial/NBRC_14.yaml	CultureMech:007476	PROPOSED	3	3 ingredients parsed	Agar15gpH 8.2*Dilute artificial seawater or natural seawater with distilled water.
+bacterial/NBRC_1401.yaml	CultureMech:007477	HOLD: parse does not round-trip	0	reassembly differs from the source; the split is ambiguous (e.g. 'Tween 800.3g' — is 80 a grade or part of the value?)	Agar (if needed)15gpH 5.6
+bacterial/NBRC_229.yaml	CultureMech:007479	PROPOSED	5	5 ingredients parsed	Agar20gpH 7.3
+bacterial/NBRC_245.yaml	CultureMech:007480	PROPOSED	5	5 ingredients parsed	Agar (if needed)20gpH 7.6
+bacterial/NBRC_26.yaml	CultureMech:007482	PROPOSED	5	5 ingredients parsed	
+bacterial/NBRC_280.yaml	CultureMech:007484	PROPOSED	4	4 ingredients parsed	Agar20gpH 7.0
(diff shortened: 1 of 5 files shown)


realmarcin:
## Review — 1 finding, addressed; central claim verified

### Verified first: is the corpus really untouched?

That is the load-bearing claim, since an earlier version of this branch wrote false chemistry to 20 records.

```
$ git diff --stat origin/main...HEAD -- data/normalized_yaml
(empty)
```

Byte-identical to `main`. The revert was complete.

Also confirmed the tool has exactly one write call (`args.out.open`, the report), all 25 records get a verdict, and none has an empty reason column.

### Finding: the tool was still named — and documented — as a repairer

Two artefacts of the reduction to report-only survived:

```
Usage::
    just repair-unparsed-compositions              # report what would change
    just repair-unparsed-compositions --apply
```

`--apply` was removed, and the recipe is `report-unparsed-compositions`. **Neither line was runnable.**

And `repair_unparsed_compositions.py` promised the opposite of what the file does. Renamed to `report_unparsed_compositions.py`.

That is not cosmetic. The whole point is that writing to these records is unsafe, and a name promising repair invites restoring the write path. The existing guard checks for `path.write_text` specifically — a differently-shaped write would slip past it — so the name carries real defensive weight.

Filed as **#180**.

### The guard added is the part worth keeping

`test_the_docstring_does_not_advertise_commands_that_do_not_exist` checks its own module: every `just <recipe>` in the docstring must exist in `project.justfile`, and every `--flag` must appear in the CLI. Verified by re-adding `--apply` and watching it fail.

It exists because this is the **third** instance this week of a docstring describing behaviour the code lacks — the retired `verify-schema-pin` reference in #106, the `--merge-audits` flag `score_review_need.py` claimed before #176, and this one. **All three were caught by review; none by tooling.**

### Also worth recording

The rebase onto current `main` hit a `project.justfile` conflict — #176 added its recipe at the same insertion point. Resolved by keeping both. That is the same-anchor hazard the goal prompt in `prompts/` warns about, hit while landing the PR that followed it.

### Verification

- [x] Corpus byte-identical to `main`
- [x] Tool has one write call, to the report only
- [x] 25 records, 14 PROPOSED / 11 HOLD, every row with a reason
- [x] New docstring guard mutation-tested
- [x] Suite: 883 passed, 2 skipped

**Does not close #166** — the records still need a curator with the NBRC catalogue.

realmarcin:
Rebased on main now that #190 has landed.

Local suite: **918 passed in 304s**, down from the >40-minute CI cancellation this
PR originally hit. The only remaining full-corpus parse is the shared session
fixture (~119s setup), which is the point of having it.

The cancellation was never a test failure — all 662 tests passed and the job was
killed at the `timeout-minutes: 40` ceiling. Cause was five tests each re-parsing
~15,900 records; fixed across #187, #190 and this branch.

Issue #170, opened 2026-08-03
The concentration baseline lives in two places and the justfile recipe doesn't enforce it


realmarcin:
## Found reviewing #169

#169 adds a baseline gate for implausible concentrations and documents the baseline in a `project.justfile` comment:

```
# Baseline history (#150): 11,540 rows across 3,914 records at the time the gate
# was added ... LOWER the baseline as records are repaired
[group('QC')]
audit-concentration-plausibility *args="":
    uv run --extra dev python scripts/audit_concentration_plausibility.py {{args}}
```

**The recipe never passes `--max-allowed`.** Running `just audit-concentration-plausibility` reports and exits 0 regardless of the count. The comment reads as though the recipe gates; it does not.

Compare the convention it claims to follow, which hardcodes the baseline into the recipe:

```
check-chebi-grounding *args:
    uv run python scripts/audit_chebi_consistency.py \
        --out reports/chebi_consistency.tsv --max-allowed 101 {{args}}
```

## Where the gate actually lives

In `tests/test_concentration_plausibility.py`, as `CONCENTRATION_BACKLOG_BASELINE = 11_540`. That is a real gate — pytest runs on every PR with no path filter — so the protection exists. It is the *documentation* that is wrong, plus a missed layer: the recipe is also absent from any workflow, so there is no non-pytest path that enforces it.

## The trap in fixing it naively

Adding `--max-allowed 11540` to the recipe puts the number in **two** places that must be lowered together as the backlog is repaired. Miss one and either the gate stops biting (justfile stale-high) or local runs fail while CI passes (justfile stale-low). That is the same drift class as #144 and #157.

## Suggested fix

Wire the recipe *and* assert the two agree, so drift is impossible:

1. `--max-allowed <baseline>` in the recipe, matching `check-chebi-grounding`.
2. A test parsing the baseline out of `project.justfile` and asserting it equals `CONCENTRATION_BACKLOG_BASELINE`.

(2) is the part that matters. Without it, two hardcoded copies of a number that is *expected to change* is a drift bug waiting to happen — and this repo has now produced four of those.

Being fixed in #169.

realmarcin:
Fixed in #169 (merged as `dd4d90ed55`), which was the PR this issue was filed from.

Verified on current `main`:

```
$ grep -A3 '^audit-concentration-plausibility' project.justfile
audit-concentration-plausibility *args="":
    uv run --extra dev python scripts/audit_concentration_plausibility.py \
[...]

Issue #168, opened 2026-08-03
Two more tracked reports are script-generated with no freshness check (#145, 5th/6th instance)


realmarcin:
## Same shape as #157, found while fixing it

`reports/chebi_consistency.tsv` (#157) was tracked, script-generated, and never checked — so it rotted 73/82 lines behind the corpus. Auditing its siblings turned up two more tracked reports with a script that writes them:

| file | writer |
|---|---|
| `reports/media_content_review_manifest.tsv` (+ `.json`) | `scripts/propose_media_variant_links.py` |
| `reports/media_growth_review_manifest.tsv` (+ `.json`) | `scripts/build_media_growth_review_html.py` |

Both are tracked. Neither has a freshness check. Neither was regenerated during the recategorizations in #115 / #120 / #137 / #143 or the composition-type restamp in #164, all of which changed records these manifests describe.

I have **not** verified whether they are actually stale — doing so means running each generator, which has side effects beyond the report (the variant-links script proposes links; the growth-review script builds HTML). That is why this is a separate issue rather than folded into #157.

## The classification question is the real one

#145 asks which derived artifacts must be refreshed. #157 established that there are at least three kinds, and the right remedy differs for each:

1. **Consumed artifacts** — indexes (#125), id registry (#144). Guard that they match the corpus; a correct committed copy has value because things resolve through them.
2. **Dated snapshots** — `reports/label_plausibility_2026-07-19.tsv`, `reports/archive/*`. Must **not** be refreshed; their old paths and counts are an accurate record of what ran.
3. **Regenerable diagnostics with no reader** — `chebi_consistency.tsv`. Should not be tracked at all; untracking removes the failure mode rather than policing it.

These two manifests need sorting into one of those buckets. `gap_fix_backlog.tsv` and `label_plausibility_after_fixes.tsv` have no writer in the repo at all, which probably makes them category 2 — but "no writer found" may just mean the generator was deleted, and that is worth confirming rather than assuming.

## Suggested first step

For each: identify whether anything reads it. That is the question that decided #157 — a diagnostic with no reader should not be tracked, while a consumed artifact should be guarded. `chebi_consistency.tsv` turned out to have zero readers across scripts, justfiles, workflows and docs.

realmarcin:
Partly addressed by #203 (merged).

`just audit-derived-artifacts` now inventories and classifies **all 93 tracked
derived artifacts**, and `--check` regenerates the freshness-checkable ones to a
[...]

Issue #121, opened 2026-07-25
deep_research_priority reports are not reproducible: prioritizer reads gitignored research/


realmarcin:
## Problem

`scripts/prioritize_deep_research_candidates.py` excludes "already-researched" media by checking for a non-dry-run meta yaml in `research/media/`. But `research/` is **gitignored** (`.gitignore:78`).

So the committed artifacts —
`data/import_tracking/reports/deep_research_priority.json`, `.md`, and `_top100.json`
— are a snapshot of **whoever last ran it, on their machine**, and cannot be reproduced by anyone else.

## How it surfaced

While fixing #116 I needed to refresh the reports (recipe `category` feeds the score, and I moved 73 records). Regenerating on this machine, with 391 local summaries in `research/media/`, changed the top-10 completely:

| | old committed top-3 | after regeneration |
|---|---|---|
| 1 | wilkins_chalgren_anaerobe_broth_n2_co2_for_dsm_15567 (92.0) | thermus_medium (87.0) |
| 2 | syntrophomonas_medium_for_syntrophospora_cellicola_19j_3 (89.0) | thermus_medium (87.0) |
| 3 | sulfolobus_medium_for_dsm_9790 (87.4) | thermus_medium (87.0) |

Almost none of that churn comes from the recategorization — it comes from local research state that is invisible to git. I reverted the regeneration in #120 rather than commit it, so the reports are currently stale for those 73 records (they still carry `category: bacterial`).

## Impact

- The reports cannot be reviewed in a PR: a diff is indistinguishable from "the author had researched more media locally".
- CI or another contributor regenerating them produces a large spurious diff.
- Report refreshes silently entangle two unrelated inputs: recipe data (tracked) and research progress (untracked).

## Options

1. Track the research-completion state in git — e.g. a small committed manifest of researched medium ids, rather than the full (large) summaries.
2. Make the exclusion explicit and reproducible: `--exclude-researched <manifest>`, defaulting to off, so the committed report is a pure function of tracked data.
3. Stop committing the generated reports and treat them as local artifacts.

(1) or (2) seems preferable — the report is genuinely useful in-repo, it just needs a reproducible input.

Found while working on #116 / #120.

PR #214, opened 2026-08-05
Gate flattened stock cocktails in CI (#150)

### LINKED ISSUE #118: Ingredient mis-normalization: trace-element stock-solution values and unit errors stored as final medium concentrations

realmarcin:
#118 asked for an audit **and** a corpus repair. #135 shipped the audit and said so
outright — *"Does NOT close #118; the corpus repair remains"* — and #118 was closed
anyway. Fifteen issues later the corpus is still wrong and nothing stops a fresh
import reintroducing the same shape.

#150 calls prevention the highest-leverage item, ahead of the repair itself,
because without it any sweep decays. This is that.

## What was actually missing

The audit already carried `--max-allowed 11540` and the recipe already passed it.
Missing were a **CI workflow to run it** and a gate on the **specific defect
shape**.

`--max-cocktails`, baselined at **579**, is the sharper of the two. A raw row count
drifts with corpus size; a flattened stock cocktail — three or more stock-strength
trace/vitamin rows with no `solutions:` block — is a specific thing an import has
done again.

## Verified by injecting one

Three trace salts at stock strength into `lb_broth`:

```
cocktails  579 -> 580     FAIL
rows     11540 -> 11544   FAIL
exit 1
```

Corpus restored, re-checked clean, exit 0.

The first attempt at that proof used values *below* the 1 G_PER_L threshold, so
only one row flagged and the gate correctly did nothing — my test data was wrong,
not the gate.

## A mistake worth recording

My first version **removed `--max-allowed` from the recipe** and put both baselines
in a new `check-` recipe — weakening an existing gate while believing I was adding
one. I had grepped for the recipe *name*, seen one line, and concluded it wasn't
wired without reading the body.

`test_justfile_baseline_matches_the_test_baseline` caught it, which is precisely
what that test exists for (#170: two copies of a number meant to change). Now one
recipe carries both flags, and the cocktail baseline has the same drift guard.

## Scope

The **3,914-record backlog is not repaired here.** The baselines hold the line
while it is worked through, per the convention `check-chebi-grounding` already
uses.

Separately worth knowing for the repair: I checked feasibility, and **183 of 216
matched cocktails have their addition volume upstream** in `data/raw/togo`
(`Trace mineral solution … 1 ml`). So the repair is recoverable from tracked data
rather than invented — which was the reason to hesitate.

Refs #150.

### DIFF
--- a/.github/workflows/concentration-plausibility.yaml
+++ b/.github/workflows/concentration-plausibility.yaml
@@ -1,0 +1,62 @@
+name: concentration-plausibility
+
+# Gate for #150: stops a fresh import re-flattening a stock solution inline.
+#
+# #118 asked for an audit AND a corpus repair; #135 shipped the audit only, and
+# nothing has since stopped the defect being reintroduced. The corpus backlog
+# (3,914 records, 579 flattened cocktails) is NOT fixed by this workflow — the
+# baselines hold the line while it is worked through, per the --max-allowed
+# convention already used by check-chebi-grounding.
+#
+# The cocktail count is the sharper of the two baselines. A raw row count drifts
+# with corpus size; a new flattened cocktail is a specific defect shape that an
+# import has reintroduced.
+
+on:
+  pull_request:
+    # Scoped to normalized_yaml: the detectors read only that tree. merge_yaml is
+    # derived from it, so a defect introduced *only* in merge_yaml would not be
+    # caught here — same caveat as chebi-consistency.
+    paths:
+      - 'data/normalized_yaml/**'
+      - 'scripts/audit_concentration_plausibility.py'
+      - 'project.justfile'
+      - '.github/workflows/concentration-plausibility.yaml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: concentration-plausibility-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  plausibility:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
+        with:
+          enable-cache: true
+
+      - name: Install just
+        uses: extractions/setup-just@v2
+
+      - name: Sync dependencies
+        run: uv sync --frozen
+
+      - name: Check concentration plausibility
+        run: just audit-concentration-plausibility
+
+      - name: Upload plausibility report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: concentration-plausibility-report
+          path: |
+            data/import_tracking/reports/concentration_plausibility.tsv
+            data/import_tracking/reports/concentration_plausibility_by_record.tsv
--- a/project.justfile
+++ b/project.justfile
@@ -225,7 +225,7 @@ audit-selective-agent-mismatch *args="":
 [group('QC')]
 audit-concentration-plausibility *args="":
     uv run --extra dev python scripts/audit_concentration_plausibility.py \
-        --max-allowed 11540 {{args}}
+        --max-allowed 11540 --max-cocktails 579 {{args}}
 
 # Merge locally-completed Edison runs (research/media/*-meta.yaml, gitignored)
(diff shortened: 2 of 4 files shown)


realmarcin:
## Review: 2 rounds, converged

**Round 1 — 1 finding, filed as #215 (out of scope here).**

Probing the `merge_yaml` caveat I had copied from `chebi-consistency.yaml`, I
found it is accurate but its consequence is larger than the comment suggests:
`data/merge_yaml/merged*/` holds **10,433 tracked media records** — real records
with CultureMech ids — and running the plausibility detectors over a 4,000-record
sample gives **5,005 implausible rows**.

So roughly half the tracked media records in the repo sit outside *every* quality
gate, on an assumption ('merge_yaml is always regenerated from normalized_yaml')
that nothing verifies. That is the shape which produced #121, #144, #157 and #168,
so it is filed rather than assumed benign. Not fixed here: #214 gates what the
detectors actually read, and the merge_yaml question is repo-wide.

Worth noting I copied that caveat without checking it applied — it happened to.

**Round 2 — 0 findings.**

- running the gate does not dirty tracked reports (0 files changed after)
- 234s runtime against `timeout-minutes: 20` (CI measured 4m33s)
- both baselines carry the #170 drift guard, with 'lower both together' guidance

Converged at 2 rounds, so no codex escalation.

**CI:** the new `plausibility` job passes at the baseline, alongside consistency,
validate and pytest.

PR #216, opened 2026-08-06
Split HYDRATE_RELAXED into three directional hydration verdicts

realmarcin:
Lands the hydration-verdict redesign that `CultureBotAI/MediaIngredientMech#242` scoped. It must land here (the hub) first because `validate_id_label_correspondence.py`, `chem_formula.py`, and `test_id_label_plausibility.py` are vendored byte-identical into the spokes; the spokes then bump `.vendored_canon_ref`.

## Why a redesign, not a relocation

`chem_formula.compare_formulas` returns `HYDRATE_RELAXED` whenever the non-water **skeleton** matches and the full counts don't. That is **symmetric**, so the obvious "label carries extra water" verdict is wrong half the time:

| label | term | reality |
|---|---|---|
| `NiCl2 x 6 H2O` | nickel dichloride (anhydrous) | label has water ✅ the wanted case |
| `MgSO4` | magnesium sulfate **hepta**hydrate | **term** has water |
| `MnSO4 x 4 H2O` | manganese(II) sulfate **hexa**hydrate | different states |

The first wants a hydrate term or `cas:` mint; the last two want a **different term**. A single verdict prescribes the wrong remedy for two of three.

## What this does

Three verdicts, chosen by the water count on each side:

- `HYDRATE_ON_ANHYDROUS_TERM` — label has water, term has none
- `ANHYDROUS_ON_HYDRATE_TERM` — term has water, label has none
- `HYDRATION_STATE_MISMATCH` — both have water, different amounts

`chem_formula.hydration_states(label, formula)` recovers `(label_waters, term_waters)` robustly: the folded **oxygen delta** between the two multisets is the water difference — correct whether the ontology formula folds waters into H/O (`H14MgO11S`) or spells them out (`Cl2Mn.4H2O`) — anchored to an absolute pair by the label's own stated count. A gap that isn't a whole number of waters (`dH ≠ 2·dO`, e.g. an –OH vs –H structural difference) is refused, so the old tolerant pass is preserved when the split can't be decided.

All three:
- start in `_WARN_VERDICTS` (non-blocking — a hydrate on its anhydrous parent is a defensible interim mapping),
- are waivable through the `exceptions` allow-list (the escape hatch #242 noted was missing),
- appear in the hardcoded summary loop (else invisible to `just validate-products`).

## Tests

`test_id_label_plausibility.py`: the three directions end-to-end through `classify`, `hydration_states`/`_label_water_units` units, and a guard that the verdicts are non-blocking. The prior `NiCl2 x 6 H2O → OK_ID_ONLY` case moved from "still passes" to `HYDRATE_ON_ANHYDROUS_TERM` — that reclassification is the point. 52 pass.

Refs `CultureBotAI/MediaIngredientMech#242`, #238.

### DIFF
--- a/scripts/chem_formula.py
+++ b/scripts/chem_formula.py
@@ -255,3 +255,56 @@ def compare_formulas(label: str, ontology_formula: str) -> str | None:
     if ws and set(ws) == set(gs):
         return "SUBSCRIPTS_LOST"
     return "CONFLICT"
+
+
+def _label_water_units(text: str) -> int:
+    """Waters of hydration written in a label string; 0 when none is stated.
+
+    Mirrors the tail/dot hydrate handling ``parse_formula`` applies, so it reads
+    the same count that folded into the element multiset: "NiCl2 x 6 H2O" -> 6,
+    "MnSO4 x H2O" -> 1 (the ``x`` is the separator, count implied), "MgSO4" -> 0.
+    A variable multiplier ("VOSO4·xH2O") never reaches here -- ``parse_formula``
+    returns None for it, so ``compare_formulas`` is None and the caller stops.
+    """
+    s = str(text or "").strip()
+    total = 0
+    m = _HYDRATE_TAIL_RE.search(s)
+    if m:
+        total += int(m.group(1) or 1)
+        s = s[: m.start()].strip().rstrip(".·・")
+    m = _HYDRATE_DOT_RE.match(s)
+    if m:
+        total += int(m.group(2) or 1)
+    return total
+
+
+def hydration_states(label: str, ontology_formula: str) -> tuple[int, int] | None:
+    """Waters of hydration on each side of a ``HYDRATE_RELAXED`` pair.
+
+    Returns ``(label_waters, term_waters)`` when the label and the ontology
+    formula share a non-water skeleton and differ ONLY by water of hydration;
+    None when that is not the case, so the caller keeps the old tolerant pass.
+
+    The term's molecular formula may fold its waters into the H/O totals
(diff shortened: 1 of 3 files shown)


realmarcin:
## Self-review (adversarial)

Probed `hydration_states` against the failure modes that matter:

- **Robust to formula representation.** Recovers `(0, 7)` for `MgSO4` whether the heptahydrate term folds its waters (`H14MgO11S`) or spells them out (`MgO4S.7H2O`) — the oxygen delta is representation-independent. Flattened, dot-notation, and label-side implied-monohydrate (`x H2O` → 1) all verified.
- **The `dH == 2·dO` water heuristic is tight, not loose.** The obvious worry is a non-water H/O difference with a coincidental 2:1 ratio being mislabeled. But with an identical non-water skeleton, a difference that is exactly H2·O1 per unit *is* water; the near-misses refuse correctly — `NaOH` vs `Na` (1:1) → None, `Mg(OH)2` vs `Mg` (2:2) → None. Not a defect.
- **No correct mapping regresses.** Same-water hydrate pairs are `MATCH` (→ OK_ID_ONLY) and never reach the split; only differing counts do.

**One rollout consideration (filed):** records previously reported `OK_ID_ONLY` for a hydrate on its anhydrous parent now surface as non-blocking warnings. Enforce is unaffected (the verdicts aren't in `_ERROR_VERDICTS`), but `just validate-products` summaries and the drift report will show new warning rows once each spoke bumps `.vendored_canon_ref`. Tracked for the sync PRs in `CultureBotAI/MediaIngredientMech#260`.

No blocking findings. Ready to merge on your go-ahead; the MIM sync PR follows once this lands on `main`.

Issue #256, opened 2026-08-14
Our groundings override MIM downstream: 6,680 divergent rows, some outright wrong compounds (MIM#138)


realmarcin:
Reported downstream as MediaIngredientMech#138 (kg-microbe loader precedence). The
investigation confirms the reporter's framing: **the hazard originates in CultureMech,
and no MIM-side fix can reach it.**

## Why a MIM fix can't reach it

kg-microbe's loader resolves an ingredient with
`best_primary([chebi_id, culturemech_term_id, mim_id, kg_microbe_node_id, cas_rn])`.
`culturemech_term_id` outranks `mim_id`, and `culturemech_term_id` is populated from
**our** `term.id`. So when MIM corrects `mim_id`, the consumer still picks our value.
MIM can only fix rows we don't have an opinion on.

## Our MIM "link" carries no MIM opinion

`mediaingredientmech_chebi_term` is effectively a self-link:

| `mediaingredientmech_chebi_term.id` vs our own `term.id` | rows |
|---|---:|
| identical | 119,577 |
| different | 554 |

That is by design — `scripts/migrate_legacy_mim_to_chebi.py` states it uses "the
ingredient's OWN CHEBI id and label" and deliberately does not import MIM's
`preferred_term`. The consequence was not intended: the field looks like corroboration
from MIM but is our own assertion echoed back, so it can never carry a correction.

1,682 legacy `mediaingredientmech_term` (`MediaIngredientMech:NNNNNN`) links also remain
across 1,506 records, pointing at a scheme MIM has retired.

## Are we consuming MIM's SSSOM? Essentially no

MIM publishes `mappings/ingredient_mappings.sssom.tsv`. Exactly one script reads it —
`scripts/audit_missing_roles.py` — and only to report role-coverage gaps. **No grounding
path consumes it.** We generate our own (`output/culturemech_chebi_mappings.sssom.tsv`)
and ground independently.

## Blast radius: 6,680 ingredient rows, 115 distinct names

Where both sides have an opinion on the same ingredient name and disagree, our value wins
downstream. It is **not** simply "MIM is right" — it runs both ways, which is why this
needs curation rather than a bulk overwrite:

| name | ours | MIM | verdict |
|---|---|---|---|
| magnesium sulfate heptahydrate (160) | CHEBI:86463 **potassium aluminium sulfate** | CHEBI:31795 magnesium sulfate heptahydrate | **ours is the wrong compound** |
[...]

PR #257, opened 2026-08-14
Correct groundings that name the wrong chemical species (#256)

realmarcin:
Closes part of #256. Reported downstream as MediaIngredientMech#138.

## The precedence hazard, confirmed

kg-microbe resolves an ingredient with `best_primary([chebi_id, culturemech_term_id, mim_id, ...])`.
`culturemech_term_id` is populated from **our** `term.id` and outranks `mim_id` — so a wrong id
here overrides MIM's correct one downstream, and **no MIM-side fix can reach it**.

Our `mediaingredientmech_chebi_term` cannot catch it either: it is a self-link to our own
`term.id` in 119,577 of 120,131 rows (by design — `migrate_legacy_mim_to_chebi.py` uses "the
ingredient's OWN CHEBI id"). It looks like corroboration from MIM but is our own assertion
echoed back. We also consume MIM's SSSOM in exactly one place (`audit_missing_roles.py`, role
coverage only) — no grounding path reads it.

## What this PR fixes

Only groundings where the record's **own evidence contradicts the id**:

| ingredient | was | which is | now |
|---|---|---|---|
| Magnesium Sulfate Heptahydrate (160) | CHEBI:86463 | **potassium aluminium sulfate** | CHEBI:31795 |
| Glucose, glucose (522) | CHEBI:42758 | **aldehydo-D-glucose** (open-chain tautomer) | CHEBI:17234 |
| D-Glucose, D-glucose, D(+)-Glucose, Dextrose (168) | CHEBI:42758 | same | CHEBI:17634 |

The magnesium rows carry `label: magnesium sulfate heptahydrate`, `CAS: 10034-99-8`, `MW: 246.47`
— every piece of the record's own evidence says CHEBI:31795. The other names on CHEBI:86463
(`AlK(SO4)2`, `Aluminum potassium sulfate`, `KAl(SO4)2`) are **correct** and untouched; the two
sets share no record, so this is a bad id rather than a row-alignment slip.

**1,434 fields across 797 records.** Both replacement ids are the ones MIM already asserts, so
this removes ~850 rows of divergence rather than trading one disagreement for another.

## Deliberately not in scope

The judgement calls from #256 — EDTA anion vs acid, thiamine vs vitamin B1, vitamin B12 vs
cyanocobalamin, and the 7 Cobalamine/Cobalamin rows blocked on MediaIngredientMech#363. Those
need curation, not a sweep. #256 tracks them.

## Two things the process caught

- **Keying.** Corrections key on `preferred_term` **and** id, never id alone — keying on the id
  would have rewritten the correctly-grounded aluminium salts. A test pins this.
- **The canary earned its keep.** The first implementation round-tripped YAML, which reflows every
  long `notes:` string: a one-id change arrived as 32 insertions / 67 deletions. Across 797 files
  that would have buried the real edit. Switched to in-place line editing; insertions now equal
  deletions (2,702 / 2,702).

## Verification

- `pytest`: 1,139 passed, 7 new tests
- residual `CHEBI:42758`: 0; residual `CHEBI:86463`: only the three aluminium names
- derived artifacts refreshed in the same commit


realmarcin:

Probed the three ways this change could be wrong. Results:

**1. Stale labels left behind.** The applier rewrites `label:` only when it follows `id:`
at the same indent; a `label:`-before-`id:` ordering would leave the old label stranded on
a new id. Checked: `aldehydo` appears **0 times** in the corpus post-change. No stragglers.

**2. Over-reach onto correctly-grounded rows.** CHEBI:86463 is legitimately carried by the
aluminium salts. Post-change it survives on exactly `AlK(SO4)2` (320), `Aluminum potassium
sulfate` (128), `KAl(SO4)2` (2) and nothing else. The independent count also reconciles:
1,884 total occurrences − 450 aluminium = 1,434 changed, matching the applier's report.

**3. Did the fix actually make the corpus consistent?** Partly — and this is the honest
limitation. Per-name it now is (`Glucose`→17234 ×2120, `D-Glucose`→17634 ×589,
`D(+)-Glucose`→17634 ×227, all single-valued). But **`Dextrose` remains split three ways**
(17234 ×157, CHEBI:4167 D-glucopyranose ×74, 17634 ×27).

That split is *pre-existing*, not introduced here — Dextrose sat on three distinct ids
before this PR (42758/17234/4167) and three after. This PR moved one group off the wrong
tautomer; it did not harmonize the rest.

## Filed, not fixed here

- #258 — glucose family split across four ids (`Dextrose` especially); needs a general
  consistency check across high-frequency ingredients, not a one-compound patch
- #259 — `term.label` carries the ingredient string rather than the ontology label
  (CHEBI:31795 labelled `MgSO4 x 7 H2O` ×6,150; 49 empty labels)

Neither blocks this PR: both are pre-existing, both are broader in scope, and #259 is
arguably the reason the wrong-compound groundings survived unnoticed — a label that
disagreed with its id would have been visible on sight.

Issue #244, opened 2026-08-07
Corpus media_term ids disagree with MediaDive: 165 cocktail records resolve to a different medium


realmarcin:
Found while extending the #150 cocktail repair to KOMODO-sourced records.

To reach a record's stock composition, its medium id is resolved to MediaDive
(`mediadive.medium:<n>` directly, or `komodo.medium:<n>` via the tracked
`komodo_dsmz_mappings.json`). Fetching that medium and comparing NAMES shows the id
frequently points at an unrelated medium:

```
KOMODO 1010  "Artificial SEAWATER MEDIUM"    -> MediaDive 1010 "DESULFOPILA CORRODENS MEDIUM"
KOMODO 294   "PELOBACTER ACIDIGALLICI MEDIUM"-> MediaDive 294  "SYNTROPHUS HQGo1 MEDIUM"
KOMODO 903   "THERMOANAEROBACTERIUM medium"  -> MediaDive 903  "CALDANAEROBIUS MEDIUM"
mediadive.medium:508  "PYRODICTIUM ABYSSI medium" -> MediaDive 508 "CARBOXYDOTHERMUS MEDIUM (HETEROTROPH)"
```

**165 of 312** resolved cocktail records disagreed by name. Note the last example
carries a **direct** `mediadive.medium:` id — so this is not only a KOMODO-mapping
artifact; some corpus MediaDive ids are themselves wrong or stale.

## Why it matters
Any tool that pulls upstream data keyed on these ids (composition backfill, stock
volumes, cross-references, KG edges) will silently import another medium's data.
The #150 applier guards against it by comparing names before use — that guard is
what surfaced this — but the underlying ids remain wrong in the corpus.

## Likely causes, not yet separated
- MediaDive renumbering since the KOMODO/MediaDive snapshots were taken.
- KOMODO's own `dsmz_medium_number` being an identity default (43.5% of that mapping
  is `komodo_id == dsmz_medium_number`) rather than a verified cross-reference.
- Genuine mis-assignment during import.

## Suggested work
1. Audit all `mediadive.medium:` / `komodo.medium:` ids in the corpus against the
   live MediaDive `/rest/media` name list, not just the cocktail subset.
2. Where the name disagrees, either re-resolve to the correct id (by name, with the
   collision caveat from #239) or mark the id unverified.
3. Consider recording a `verified_against` datestamp so a future renumbering is
   detectable rather than silent.

[...]

Issue #221, opened 2026-08-06
Cocktail baseline lacks the anti-vacuous check the row baseline has (#214 review)


realmarcin:
From review of #214.

`tests/test_concentration_plausibility.py::test_the_gate_baselines_match_the_current_corpus`
guards the **row** baseline against becoming vacuous:

```python
assert rows_baseline - len(rows) <= 200, (
    f"gate baseline {rows_baseline} is {rows_baseline - len(rows)} above the "
    "actual count — tighten it or it will never fire")
```

but the **cocktail** baseline — which #214 itself calls "the sharper of the two"
— gets only:

```python
assert COCKTAIL_BASELINE > 0
```

So as the 579-record cocktail backlog is repaired under #150, `COCKTAIL_BASELINE`
can drift far above the real count and the sharper gate silently stops being able
to fire, with no test noticing. The row baseline's ≤200 check partially masks this
(repairing cocktails also drops rows), but that is incidental, not a guarantee.

## Fix
Add a symmetric anti-vacuous assertion for the cocktail baseline: compute the
current flattened-cocktail count and assert `COCKTAIL_BASELINE - actual <= tol`.
Requires counting cocktails from parsed records in the test (the script does it as
`sum(1 for s in summary if s["flattened_cocktail"] == "yes")`).

Low severity: current baseline (579) equals the actual count, so nothing is vacuous
today. Filed rather than folded into #214 because it hardens the gate's self-test
rather than fixing a defect in what #214 ships.

PR #261, opened 2026-08-15
Resolve KOMODO cocktails through their base medium (#150)

realmarcin:
Part of #150. Takes flattened cocktails **279 → 185** (#150 now 66% repaired from the
original 579); implausible rows **10,244 → 9,750**.

## The blocker was addressing, not research

111 of the 119 blocked records whose stock we can already name carry a `komodo.medium:`
id, and `apply_cocktail_nesting.source_medium` refuses those **on purpose**: `KOMODO_294`
stamps "DSMZ Medium: 294" on a record whose DSMZ 294 is a different medium (#244). So the
stock was known, the volume existed upstream, and nothing connected them.

The KOMODO key carries the medium number, and it can be derived **twice** — structurally
(`142_12346` → `142`) and from the tracked KOMODO→DSMZ export. The new fetch step requires
both to agree before it fetches anything. That is a real check: a disagreement between a
stamped number and a derived one is precisely what produced #244.

## Only one evidence class is an assertion

A stock's addition volume is a property of the **citing medium**, and this run demonstrates
it flatly: MediaDive medium 142 adds `Trace element solution (Vishniac & Santer)` at
**0.2 ml**; medium 69 adds the same stock at **5 ml**. 25× apart.

| class | when | records | written to |
|---|---|---:|---|
| `READ_FROM_THIS_MEDIUM` | bare medium number **and** the fetched medium's name agrees | 10 | `concentration` |
| `CROSS_MEDIUM_INFERENCE` | variant of a base medium (own recipe never published) | 76 | `concentration_candidates` |
| `CROSS_MEDIUM_INFERENCE` | bare id whose name disagrees — the #244 signature | 41 | `concentration_candidates` |

Candidates carry both `support` and `counterevidence` and leave `concentration` unset, per
the mechanism added in #255. **`KOMODO_294` lands in the third class**, with counterevidence
naming both media — the outcome the design wanted for the record that motivated the guard:

> The fetched medium is named 'SYNTROPHUS HQGö1 MEDIUM' but this record is 'PELOBACTER
> ACIDIGALLICI MEDIUM'; the identification rests on the medium number alone.

Of the 94 records actually nested, all 181 solutions got candidates and none got an
assertion — the 10 assertable records were independently refused by the existing
refuse-partial guard. A test pins the assert path so that stays selection, not silent
breakage.

Every move still passes the **name+value** match against the stock composition, so nothing
rests on the medium resolution alone.

## Also fixes a latent aliasing bug

`split_parts` used `dict()` — a shallow copy — so a split row shared its nested `term` and
`mediaingredientmech_chebi_term` dicts with the ingredient it came from, and the two ended
up in **different solutions**. `yaml.dump` emitted them as `&id001`/`*id001`: two solutions
aliasing one object, where mutating either changes both. Now `deepcopy`. Caught by reading
the canary's diff.

34 records written before this fix still carry anchors — left for a follow-up rather than
mixed into this diff.

## Verification

- `pytest`: 1,158 passed, 12 new
- gate baselines tightened to the new counts in **both** the test and the justfile (the
  anti-vacuous guard from #221 fired correctly, twice)
- indexes and derived artifacts regenerated in the same commit
- `komodo_base_volumes.json` left untracked, matching the sibling MediaDive cache


realmarcin:

Probed four ways this could be wrong.

**1. Implausible candidate volumes.** Distribution across all 181 candidates written:
`{0.2: 6, 0.5: 1, 1.0: 165, 5.0: 6, 10.0: 3}` ml/l. Nothing outside 0.01–50. The six
`0.2 ml` entries are the medium-142 variants correctly taking *that medium's* value
rather than the corpus mode of 1 ml — the per-medium reading working as intended.

**2. Components left in two places.** A half-repair (row moved into a solution but also
still in `ingredients:`) would be worse than no repair. Checked all 94: **0 records** have
a component name in both `ingredients` and a `solutions[].composition`.

**3. Is the assert path actually reachable?** All 94 nested records got candidates and
none got an assertion, which looked like a broken branch. It isn't — the 10 assertable
records were independently refused by the pre-existing refuse-partial guard and remain
flat. `test_asserted_basis_writes_concentration` pins the branch so a future regression
shows up as a test failure rather than as silence.

**4. New YAML anchors.** The `deepcopy` fix holds: anchor-carrying records stayed at 34
(all pre-existing) across a 94-record apply.

## Filed, not fixed here

- #262 — the 41 name-mismatch candidates mix genuine upstream renames (`69`:
  Thiobacillus novellus → Starkeya novella, volume probably assertable) with genuine
  #244 collisions (`294`: Pelobacter vs Syntrophus, candidate probably wrong and should
  be withdrawn). Distinguishing them needs a taxonomy check, not string similarity.
- #263 — the 34 records still carrying `&id001`/`*id001` from before the deepcopy fix.
- #264 — `identify_cocktail_stocks` reports `invariant: yes` on n=1 and n=2 samples. This
  PR's own data falsifies one of those rows: the Vishniac stock is reported invariant at
  5 ml on n=2, and actually ranges 0.2–5 ml.

None blocks this PR. #262 is the one I'd take next — it can both promote and withdraw
candidates this PR deliberately left conservative.

Issue #260, opened 2026-08-14
Resolve ingredient groundings from MIM's published label index instead of local term.id


realmarcin:
Filed from the MediaIngredientMech side, where MediaIngredientMech#138 has been fixing CultureMech/MIM grounding divergences one row at a time since it was opened. The divergences are systemic, and there is a published artifact that retires the class.

## The measurement

Joining CultureMech's 3,916 distinct ingredient names against MIM:

| | count |
|---|---|
| names with a CultureMech `term.id` | 1,584 |
| names MIM can resolve | 2,112 |
| both have an opinion | 1,569 |
| — agree | 1,501 |
| — **disagree** | **68** (3,386 occurrences) |
| CultureMech has a term, MIM does not | 59 |
| **MIM resolves, CultureMech has no term** | **543** |

So MIM already resolves *more* of CultureMech's own ingredient names than CultureMech does (2,112 vs 1,584).

## The 68 disagreements favour MIM everywhere I checked

| ingredient | occ | CultureMech | MIM | which is right |
|---|---|---|---|---|
| `EDTA` | 520 | `CHEBI:64755` EDTA(2−) | `CHEBI:4735` free acid | MIM — ChEBI defines the (2−) as deprotonation of *two of four* carboxy groups; the label asserts no charge |
| `4-Aminobenzoic acid` | 141 | `CHEBI:194474` **4-ammoniobenzoate** | `CHEBI:30753` the acid | MIM — a zwitterion; CAS `150-13-0` and 1,968 MIM occurrences say the neutral acid |
| `Ca-pantothenate` | 137 | `CHEBI:29032` bare anion | `CHEBI:31345` calcium salt | MIM |
| `Beef heart` | 164 | `UBERON:0000948` (the organ) | `FOODON:00004410` | MIM — a food product, not an anatomical structure |
| `Calf brains` | 164 | `UBERON:0000955` (the organ) | `UNMAPPED_0170` | MIM — honestly unmapped |
| `Infusion from Potatoes` | 1 | `FOODON:03316428` **sweets** | `UNMAPPED_0251` | MIM — and CultureMech stores `label: 'potato extract'` against that id, which FOODON calls `sweets` |

Note the last two rows are exactly MediaIngredientMech#138's remaining items, and `4-Aminobenzoic acid` alone is **251 entries** across 251 files here, every one a PABA label with an empty `term.label`.

## The proposal

[...]

Issue #259, opened 2026-08-14
term.label often carries the ingredient string instead of the ontology label


realmarcin:
Found reviewing #257; noted in passing in #256, filed separately because it is a
distinct and mechanically-checkable class.

`term.label` is expected to be the ontology term's label. In much of the corpus it is
the ingredient string from the source recipe instead:

| id | label as stored | rows | actual CHEBI label |
|---|---|---:|---|
| CHEBI:31795 | `MgSO4 x 7 H2O` | 6,150 | magnesium sulfate heptahydrate |
| CHEBI:31795 | `Magnesium sulphate heptahydrate` | 1 | (spelling variant) |
| CHEBI:31795 | `` (empty) | 49 | — |
| CHEBI:17439 | `Cyanocobalamin` | 23 | cyanocob(III)alamin |

The same id therefore appears under many different labels, and both forms coexist for
the same term (CHEBI:17439 carries its real label in 156 rows and the ingredient string
in 23).

Consequences:
- Label-based review and diffing is defeated — a reviewer cannot eyeball whether an id
  is right, which is exactly the check that would have caught the wrong-compound
  groundings fixed in #257.
- Empty labels (49 on CHEBI:31795 alone) carry no signal at all.

The ingredient string is already preserved in `preferred_term`, so `term.label` can be
refilled from the ontology without losing anything. Suggested: a checked artifact that
refills `term.label` from CHEBI and fails CI when an id and its label disagree.

PR #266, opened 2026-08-15
Ground PABA to the neutral acid, drop the bogus potato mapping (#260)

realmarcin:
Addresses the tactical items in #260, reported from MediaIngredientMech#138. Both are
cases where **our** value wins downstream — kg-microbe resolves with
`best_primary([chebi_id, culturemech_term_id, mim_id, ...])` and `culturemech_term_id`
comes from our `term.id` — so no MIM-side fix can reach them.

## PABA — 251 records

`CHEBI:194474` is **4-ammoniobenzoate**, the zwitterion. The records' own evidence
contradicts it: every row carries `CAS: 150-13-0` (the neutral acid, `CHEBI:30753`) in
its notes, and every row has an **empty** `term.label` — so nothing on the record ever
asserted the zwitterion. The label is now filled rather than left blank.

| name | rows |
|---|---:|
| `4-Aminobenzoic acid` | 237 |
| `p-Amino Benzoic Acid` | 12 |
| `p-amino benzoic acid` | 2 |

`CHEBI:194474` carries no other ingredient name in either corpus, so unlike the
CHEBI:86463 case in #257 there is nothing here to preserve — but the correction is still
keyed on name **and** id, per that lesson.

## Infusion from Potatoes — 2 records

Grounded to `FOODON:03316428`, whose actual label is **`sweets`**. The stored label said
`potato extract`, which is not that term's label and never was. Removed rather than
reground — MIM records this ingredient as `UNMAPPED_0251`, and honestly unmapped beats
confidently wrong. The dangling `Mapping: manual (confidence: 0.85)` note is replaced
with the reason, so the empty grounding reads as deliberate rather than as an oversight.

## The actual finding: merge_yaml was stale

`data/merge_yaml/merged` is a **derived product** of `normalized_yaml`, and PR #257 only
ran against `normalized_yaml`. So after #257 merged, the merged corpus still carried:

| id | is | merged files still affected |
|---|---|---:|
| `CHEBI:42758` | aldehydo-D-glucose | 608 |
| `CHEBI:86463` | potassium aluminium sulfate (on magnesium sulfate heptahydrate) | 118 rows |

Corrected here as well — **1,359 rows across 717 merged records**. This is why the PABA
count is 251 rather than the 142 in `normalized_yaml` alone. Filed as a follow-up: a
corpus fix that reaches one corpus and not its derived twin will recur.

## Investigated, deliberately not in scope

`TOGO_M2314_LB_medium.yaml` also matches `FOODON:03316428`, but in **prose**:

> Original FOODON mapping was FOODON:03316428 tryptone.

That is a provenance note recording a superseded mapping, not a live grounding — nothing
resolves through it, and rewriting it would edit history. Worth noting it is a *third*
distinct label for that one id (`sweets` / `potato extract` / `tryptone`), and `tryptone`
is wrong too.

## Verification

- `pytest`: 1,147 passed, 1 new (this branch is off `main`, so it does not include #261's 12 tests)
- residual `CHEBI:194474`: **0** in both corpora; residual `CHEBI:42758`: **0** in both
- canary first: one record, 2-line diff, correct term and filled label, before fan-out
- indexes and derived artifacts regenerated in the same commit

Leaves the structural proposal in #260 — consuming MIM's `label_index.csv` so this class
stops recurring — open and untouched.


realmarcin:

**1. Is the diff pure substitution?** 874 files, **2,898 insertions / 2,900 deletions**.
The −2 is fully accounted for: each of the two potato records drops a 4-line
`term:`/`id:`/`label:`/`notes:` block and gains a 3-line note. No reflow anywhere else.

**2. Did the correction over-reach?** `CHEBI:194474` carried only the three PABA name
variants in either corpus, so unlike the CHEBI:86463 case in #257 there was nothing to
preserve — but it is still keyed on name **and** id. Residual `CHEBI:194474`: **0** in
both corpora. Residual `CHEBI:42758`: **0** in both.

**3. Are the two corpora actually in sync now?** This is the check that matters, since
the PR claims to have closed a gap #257 left. Comparing modal `term.id` per ingredient
name across both roots:

```
names in both corpora: 1571;  modal grounding DISAGREES on: 0
```

**4. Did removing the potato grounding damage the record?** The ingredient is still
present with its concentration intact (`4 G_PER_L`); only `term` is gone, and the note now
explains why. Asserted in the apply script rather than checked by eye.

## Filed, not fixed here

- #267 — the underlying process gap. `data/merge_yaml/merged` is a derived product that is
  tracked but not regenerated by `just refresh-derived`, so a corpus fix run against
  `normalized_yaml` silently fixes half the corpus. That is how #257 left 608 merged files
  on aldehydo-D-glucose after it merged. The cheapest durable fix is a CI check asserting
  what item 3 above verified by hand — it would have caught #257's gap the day it landed.

## One thing to flag rather than fix

`TOGO_M2314_LB_medium.yaml` carries `FOODON:03316428` in a prose provenance note calling it
`tryptone`. It is not a grounding and nothing resolves through it, so I left it. But that
makes **three** distinct labels recorded for one id across this repo — `sweets` (the real
one), `potato extract`, and `tryptone` — which is a small illustration of why #259
(term.label carrying the ingredient string) is worth doing.

PR #268, opened 2026-08-15
Let composition, not the medium name, decide a KOMODO volume (#262)

### LINKED ISSUE #262: Triage the 41 KOMODO candidates whose medium name disagrees: renames vs #244 collisions
Found reviewing #261.

#261 resolves KOMODO-sourced cocktails through their base medium number and classifies
41 records as `CROSS_MEDIUM_INFERENCE` because the fetched medium's name disagrees with
the record's. That class currently mixes two very different situations:

**Genuine upstream rename** — the medium is right, the name moved:
`komodo.medium:69` is `THIOBACILLUS NOVELLUS medium` here and `STARKEYA NOVELLA MEDIUM`
in MediaDive. Same organism; *Thiobacillus novellus* was reclassified as *Starkeya
novella*. The volume could legitimately be asserted.

**Genuine #244 collision** — the medium number means different things on each side:
`komodo.medium:294` is `PELOBACTER ACIDIGALLICI MEDIUM` here and `SYNTROPHUS HQGö1
MEDIUM` upstream. Different medium. The proposed volume is probably **wrong**, not merely
unverified, and the candidate should be withdrawn rather than left for someone to promote.

#261 treats both conservatively (candidate, never assertion, counterevidence naming both
media), which is safe but leaves the wrong ones sitting in the corpus looking like open
proposals.

Triage needs a taxonomy check rather than string similarity.
`scripts/audit_mediadive_ids.py` already carries `KNOWN_UPSTREAM_RENAMES`, and
`src/culturemech/taxonomy/similarity.py` exists — between them a rename should be
distinguishable from a collision.

Proposed outcomes:
- rename confirmed → promote to `READ_FROM_THIS_MEDIUM`
- collision confirmed → delete the candidate and report the record as unresolvable
- neither → leave as-is

Note the nesting itself is not in question in either case: every moved row matched the
stock composition on name AND value, so those rows are flattened stock entries regardless
of which medium supplied the volume. Only the volume is at stake.

realmarcin:
Closes #262.

## The premise in #261 was too pessimistic

#261 refused to assert a KOMODO record's addition volume whenever the fetched medium's
**name** disagreed with the record's, saying the identification "rests on the medium
number alone". 41 records landed in that bucket. It doesn't — the corpus holds the
evidence to settle it.

## Names can't discriminate; values can

| comparison | unrelated media score | same medium |
|---|---|---|
| compound **names** | ~0.97 | ~0.97 |
| compound **values** | **0.00** | **1.00** |

Anaerobic DSMZ media share a backbone (salts, trace elements, vitamins, bicarbonate,
sulfide), so a name-set comparison scores 0.97 between media that have nothing to do with
each other. I checked this before trusting it: each record scored against its own resolved
medium **and** against three unrelated media. Own medium 1.00 over 16–32 shared compounds;
unrelated media 0.00 every time.

Across the 41 the split is clean, with a wide gap and nothing inside it:

- **23 nested records at exactly 1.00** (15–47 shared compounds)
- **8 at 0.70–0.96**

So the name disagreements are overwhelmingly **upstream renames**, not #244 collisions:
*Thiobacillus novellus* → *Starkeya novella*, *Clostridium pfennigii* → *Oxobacter
pfennigii*, *Thiobacillus aquaesulis* → *Annwoodia*.

**KOMODO 294 — the record that motivated the whole guard** (`PELOBACTER ACIDIGALLICI`
here, `SYNTROPHUS HQGö1` upstream) — reproduces medium 294 across **30 of 30 compounds at
identical values**. Same medium, different name.

## Independent corroboration

11 of the 23 already carry a `variant_children` / `parent_media` link naming the very
record MediaDive titles that medium. `pyrodictium_abyssi_medium` is recorded as a
`SOURCE_DUPLICATE` of `carboxydothermus_medium_heterotroph` — exactly MediaDive's name for
medium 508. The corpus already knew; nothing had joined it up.

Of the 12 without such a link, most are recognizable reclassifications; the remainder are
DSMZ reusing one medium for a different type strain, which doesn't affect the volume —
composition identity is what the volume attaches to.

## What changed

| | records | effect |
|---|---:|---|
| **promoted** | 23 | volume moved from `concentration_candidates` to `concentration`, preparation note stating what identified the medium |
| **sharpened** | 8 | candidate kept, but counterevidence now cites the measured disagreement ("only 89% of the 37 shared compounds carry the same value") instead of a name mismatch |

Promotion **never overwrites** a stated `concentration` — it only fills a gap. A test pins
that, along with the rule that too few shared compounds cannot identify a medium even at
perfect agreement.

## Verification

- corpus now: **496 asserted** solution volumes, **139 candidates**, **0** carrying both
- `pytest`: 1,175 passed, 8 new
- canary first: one record, inspected diff, before fan-out
- the classifier in `fetch_komodo_base_volumes.py` now uses composition too, so future
  runs classify correctly at source rather than needing this back-fill


realmarcin:

**1. Is the composition match an artifact of shared stocks?** This is the strongest
objection to the PR: two unrelated media that both add SL-10 and Seven vitamins share ~14
stock compounds at identical values for free, since a stock's composition is the same
wherever it is used. If the agreement came only from those, the promotions would rest on
nothing.

It doesn't. Every one of the 23 promoted records also matches on the medium's **main
solution**:

| main-solution compounds shared | records |
|---:|---:|
| 4 | 1 (`sypc_medium`, medium 1188) |
| 5–7 | 7 |
| 8–13 | 15 |

`sypc_medium` at 4 is the thinnest in the set. Filed as #269 — the code should *require* a
main-solution minimum rather than rely on it happening to hold.

**2. Did the metric get a control?** Yes, before I trusted it: each record scored against
its own resolved medium **and** three unrelated media. Own 1.00 over 16–32 compounds;
unrelated **0.00** every time. Without that control the 0.97 name-similarity score would
have looked like confirmation.

**3. Did promotion overwrite anything?** No solution in the corpus carries both a
`concentration` and `concentration_candidates` (0 of 635). `promote_doc` skips any
solution that already has a `concentration`, pinned by
`test_promotion_never_overwrites_a_stated_concentration`.

**4. Are the asserted volumes sane?** 496 asserted / 139 candidate solutions corpus-wide,
all within 0.05–10 ml/l.

## Filed, not fixed here

- #269 — require a minimum of *main-solution* compounds, not just total shared. Also
  re-check `sypc_medium` by hand; if the rule lands at 5 it should be demoted back to a
  candidate.

One correction to #262's framing, which I wrote: it assumed the 41 mixed renames with
#244 collisions roughly evenly and would need NCBI taxonomy to separate them. Composition
did it without taxonomy, and the answer was lopsided — 23 renames, 0 confirmed collisions,
8 genuinely-different media that stay candidates.

Issue #258, opened 2026-08-14
One ingredient name, several CHEBI ids: the glucose family is split four ways


realmarcin:
Found reviewing #257. Pre-existing — #257 did not create it (Dextrose sat on three
distinct ids before that PR and three after) — but the review made it visible.

`Dextrose` is grounded three different ways in the corpus today:

| name | id | label | rows |
|---|---|---|---:|
| Dextrose | CHEBI:17234 | glucose | 157 |
| Dextrose | CHEBI:4167 | D-glucopyranose | 74 |
| Dextrose | CHEBI:17634 | D-glucose | 27 |

Dextrose is, by definition, D-glucose, so CHEBI:17634 is the defensible target and the
other 231 rows should move. `alpha-D-Glucose` (CHEBI:17925) and `beta-D-Glucose`
(ungrounded, 16 rows) are genuinely distinct anomers and should stay distinct.

Also ungrounded in the family: `beta-D-Glucose` (16), `D--Glucose` (6, and that double
hyphen looks like a normalization artifact worth chasing separately),
`Dextrose (BD 215530)` (3), `α-D-Glucose monohydrate` (2).

This matters more than tidiness: `culturemech_term_id` outranks `mim_id` downstream
(#256), so each variant propagates as a different compound into kg-microbe, and a
consumer counting glucose-containing media gets the wrong answer.

Worth checking whether the same split affects other high-frequency ingredients before
fixing glucose specifically — the fix should be a general consistency check, not a
one-compound patch.

PR #270, opened 2026-08-15
Move Cysteine-HCl off a fluorescent dye and onto the right salt

realmarcin:
Reported from the kg-microbe session as the one item the #260 work did not cover. Both
defects are the class #257 established: **the record's own name contradicts its id.**

## Two defects

**1. `Cysteine-HCl` → `CHEBI:52891` = `QSY9 succinimidyl ester(1+)`** — a fluorescence
quencher dye. 163 rows. Not a near-miss between related species; a completely different
kind of molecule, the same shape as magnesium sulfate heptahydrate grounded to potassium
aluminium sulfate in #257.

The same string is already grounded to `CHEBI:91247` (L-cysteine hydrochloride) in 40
other rows, and the name carries no hydrate marker → the anhydrous salt is the target.

**2. Hydrate-and-HCl names → `CHEBI:17561` = plain `L-cysteine`** — neither the salt nor
the hydrate. 126 rows across `L-Cysteine-HCl x H2O`, `Cysteine-HCl x H2O`,
`Cysteine-HCl·H2O`, `L-cysteine-HCL x H2O`, `L-Cysteine-HCl·H2O`.

The corpus already uses `CHEBI:91248` (L-cysteine hydrochloride hydrate) for this
substance in **1,901 rows**, so this is the corpus disagreeing with itself — not a
judgement call about hydrate specificity of the kind #256 deliberately leaves alone.

## What deliberately does not move

The **169 rows named `L-Cysteine` / `Cysteine`** keep `CHEBI:17561`, which is correct for
the free amino acid. Only names carrying `HCl` move. Keying on the id alone would have
wrecked every one of them — the same trap as `AlK(SO4)2` legitimately holding CHEBI:86463
in #257. A test pins it.

## Scope

**399 fields across 289 records** = 289 groundings + 110 `mediaingredientmech_chebi_term`
self-links, which move with the term rather than being left contradictory. Reconciled
against an independent scan before applying.

Run against **both** corpora (`normalized_yaml` and `merge_yaml/merged`), so the
derived-corpus check added in #266 stays green.

## Verification

- residual `CHEBI:52891`: **0**; residual HCl-named rows on `CHEBI:17561`: **0**;
  legitimate plain-cysteine rows on `CHEBI:17561`: **169**, untouched
- `pytest`: 1,179 passed, 3 new
- canary first: one record, 2-line diff, before fan-out

## Note on the reported blocker

The kg-microbe session reported CultureMech sitting on `fix/paba-and-potato-groundings-260`
with 876 uncommitted files. That snapshot is stale — the work was committed and merged as
**#266** (`4058ceee32`), the branch is deleted, and `main` is clean and synced. The PABA,
D-glucose, Infusion-from-Potatoes and CHEBI:86463 fixes they listed are all on `main`
already. Details in the PR discussion.


realmarcin:

**1. Is `CHEBI:91247` right for an unmarked `Cysteine-HCl`?** This was the judgement call
I was least sure of — commercial cysteine HCl is usually supplied as the monohydrate, so
`91248` was arguable. **MIM settles it independently**: its `UNIFIED_INGREDIENT_MAPPING`
grounds both `Cysteine-HCl` and `cysteine-HCl` to `CHEBI:91247`, the same value this PR
picks, arrived at from the other side. It also confirms the defect — MIM's
`culturemech_term_id` for those rows still reads `CHEBI:52891`, the dye, because that
column is populated from our value.

**2. Did the correction over-reach?** Residual `CHEBI:52891`: **0**. HCl-named rows still
on `CHEBI:17561`: **0**. Plain `L-Cysteine`/`Cysteine` rows still on `CHEBI:17561`:
**169**, untouched. Keying on the id alone would have taken all 169.

**3. Does the count reconcile?** The script reports 399 fields; an independent scan finds
289 groundings + 110 `mediaingredientmech_chebi_term` self-links = 399. Same
field-vs-row distinction reconciled in #257.

**4. Does the derived corpus stay in step?** Run against both roots, and
`test_corpus_grounding_agreement` (added in #266) passes — which is the first time that
check has done its job on a new fix rather than retroactively.

## Filed, not fixed here

- #271 — MIM grounds `L-Cysteine HCl x H2O` (space) to CHEBI:91248 but
  `L-Cysteine-HCl x H2O` (hyphen) to CHEBI:17561. Same substance, split by punctuation.
  We had the identical split; this PR fixes our side, which leaves **63 rows** where our
  `culturemech_term_id` now disagrees with MIM's `mim_id`. Ours is the correct one and
  wins downstream, but MIM's copy stays wrong for anyone reading MIM directly. That is a
  MIM-side edit and I have not filed it there.

The general cure for both is a normalization step that collapses punctuation before
grounding — same disease as #258.

Issue #265, opened 2026-08-15
Ingredient synonyms are modelled but effectively unpopulated — 22 distinct strings corpus-wide


realmarcin:
Found while investigating whether CultureMech could resolve its own ingredient name variants, as an alternative to consuming MIM's label index (#260).

**It cannot, and the reason is data rather than schema.**

## The schema supports synonyms

`src/culturemech/schema/culturemech.yaml` defines `IngredientDescriptor.synonyms`, an `IngredientSynonym` class with `synonym_text` / `synonym_type`, and `MediaRecipe.synonyms`. The modelling is there and is fine.

## The data does not

Measured across `data/normalized_yaml/**` (15,407 media files):

| | count |
|---|---|
| ingredient entries | 169,573 |
| — carrying any synonym | **3,130 (1.8%)** |
| distinct ingredient names | 3,928 |
| — that *ever* carry a synonym | **11 (0.3%)** |
| **distinct synonym strings in the entire corpus** | **22** |

The 22 are a handful of hardcoded pairs repeated across thousands of media:

    NaCl (783)  ·  table salt (783)  ·  casein peptone (626)
    dextrose (621)  ·  D-glucose (621)  ·  tryptone (464)

So eleven ingredients have synonyms and 3,917 do not. For comparison, MIM publishes **5,656 synonyms** over 2,841 preferred terms in `docs/data/label_index.csv`.

## Why it matters

Ingredient names in this corpus vary heavily in surface form — `MgSO4·7H2O` / `MgSO4 x 7 H2O` / `magnesium sulfate heptahydrate`, `Glucose` / `D-Glucose` / `Dextrose`, `KOH` / `Potassium hydroxide`. Without synonyms, each variant is a separate unjoinable string, which is why:

- the same substance appears many times under different names with independently-assigned `term.id` values that then disagree (68 such disagreements measured against MIM in #260)
- half of CultureMech's names cannot be resolved from MIM's SSSOM either, because SSSOM also publishes only preferred terms — the join has to go through a synonym-bearing index

## Two ways forward, not exclusive

**1. Consume MIM's `label_index.csv` (#260).** It already carries the 5,656 synonyms and resolves 2,215 of CultureMech's 3,928 names. This is the cheap path and needs no data entry here.

**2. Populate `synonyms` from that same index**, so CultureMech records become self-describing rather than depending on a join at read time. This is the more durable path but is a bulk data change.

[...]

PR #272, opened 2026-08-15
Collapse ingredient names grounded several ways (#258)

### LINKED ISSUE #258: One ingredient name, several CHEBI ids: the glucose family is split four ways

realmarcin:
Closes #258.

## The split was general, as #258 suspected

#258 was filed about Dextrose but asked, correctly, whether to check the general case
first. **63 of 1,583 grounded names carried more than one `term.id`, across 13,657 rows.**
Dextrose was tenth by row count.

Adds `audit_grounding_consistency.py`, which reports every such name and separates them by
*why* the ids compete — the kinds need different handling:

| kind | meaning | fixable? |
|---|---|---|
| **HYDRATE** | same substance, different waters of crystallisation | yes — the name settles it |
| **ION** | a bare ion/zwitterion where the name states the salt or neutral compound | yes |
| **OTHER** | different compounds, anomers, vitamer classes | curation, not a sweep |

## The largest correction inverts the majority

**1,161 rows named `CoSO4 x 7 H2O` sat on the ANHYDROUS cobalt sulfate; only 5 on the
heptahydrate — and the 5 were right.** A majority-wins rule would have entrenched the
error. Nothing here is decided by frequency; the ingredient's own name decides.

Selected corrections (3,985 rows in `normalized_yaml`, the same again in `merge_yaml`):

| name | was | is now |
|---|---|---|
| `CoSO4 x 7 H2O` +3 spellings | CHEBI:53470 cobalt(2+) sulfate | CHEBI:91244 heptahydrate |
| `Na2MoO4·2H2O` +3 spellings | CHEBI:75215 anhydrous | CHEBI:75213 dihydrate |
| `Na2SeO3 x 5 H2O` +3 | CHEBI:48843 anhydrous | CHEBI:131361 pentahydrate |
| `MgCl2 x 6 H2O` +2 | CHEBI:6636 anhydrous | CHEBI:86345 hexahydrate |
| `Starch` | CHEBI:85248 **gellan gum** | CHEBI:28017 starch |
| `KNO3` | CHEBI:17632 **nitrate ion** | CHEBI:63043 potassium nitrate |
| `Glycine` | CHEBI:57305 zwitterion | CHEBI:15428 glycine |
| `Dextrose` (both wrong ids) | CHEBI:17234 / CHEBI:4167 | CHEBI:17634 D-glucose |

## What deliberately does not move

Every correction keys on the exact name **and** the id, because the same "wrong" id is
*correct* for the unmarked name:

- `Na2MoO4` (69) and `Sodium molybdate` (311) stay anhydrous on CHEBI:75215
- `Gelrite` (24) and `Gellan Gum` (7) keep CHEBI:85248
- `sodium citrate dihydrate` (40) keeps CHEBI:32142
- `CoSO4`, `MgSO4`, `Na2SeO3`, `FeSO4`, `MgCl2` all keep their anhydrous ids

Verified after applying: every expected survivor is still in place.

Also untouched: **`Vitamin B12`** (cyanocobalamin vs the vitamer class — the judgement call
#256 defers, and now the largest remaining split at 2,344 rows), `Maltose` (generic vs
alpha anomer), and rows naming hydrates that don't exist (`MgCl2 x 7 H2O`).

## Result

Splits go **63 names / 13,657 rows → 50 names / 4,172 rows**. Of what remains, 6 HYDRATE
and 9 ION are mechanical follow-ups; 35 need curation.

## A bug worth recording

The audit's HYDRATE classifier stripped `hydrate` but left `hepta`, so the two labels never
shared a stem and the class never fired. **The first run reported zero HYDRATE splits,
which reads like success.** Caught by a test written against a case I already knew was a
hydrate split — not by reading the output, which looked fine.

## Verification

- `pytest`: 1,185 passed, 9 new
- both corpora, so the derived-corpus check from #266 stays green
- canary first: one record, 2-line diff, before fan-out


realmarcin:

**1. Pure substitution, or did something get added or lost?** 3,421 files, **13,810
insertions / 13,810 deletions** — exactly balanced, so no reflow and no structural change.
Independently: `    id: CHEBI:` lines number **296,167 before and 296,167 after**. No
ingredient gained or lost a grounding.

**2. Did the name+id keying hold?** This is the one that matters, because the same "wrong"
id is *correct* for the unmarked name. Post-apply, all expected survivors are in place:
`Na2MoO4` (69) and `Sodium molybdate` (311) still anhydrous on CHEBI:75215; `Gelrite` (24)
and `Gellan Gum` (7) still on CHEBI:85248; `sodium citrate dihydrate` (40) still on
CHEBI:32142; `CoSO4`, `MgSO4`, `Na2SeO3`, `FeSO4`, `MgCl2` all still anhydrous.

**3. Derived corpus in step?** `test_corpus_grounding_agreement` (from #266) passes — the
check fired on a new fix rather than retroactively, which is what it was for.

## Worth flagging rather than filing

The classifier bug in this PR is the interesting failure. It stripped `hydrate` but left
`hepta`, so labels never shared a stem and **HYDRATE reported zero splits** — output that
looks like a clean result. I only caught it because I wrote a test against a case I already
knew was a hydrate split. A report that says "no problems found" is indistinguishable from
a broken detector unless something independently asserts the detector works; the
`test_the_scan_actually_finds_groundings`-style vacuity guard added in #266 exists for the
same reason and should probably be standard for every audit script here.

No new issues from this review. The remaining 50 splits are enumerated in
`grounding_consistency.tsv`: 6 HYDRATE + 9 ION are mechanical follow-ups, and `Vitamin B12`
(2,344 rows) is the largest curation item, still deferred per #256.

realmarcin:
Cross-checked this PR's chosen `term.id` against MediaIngredientMech's answer for every name it changes. **27 names change; 11 agree with MIM, 16 disagree.** Most of the disagreement is MIM's fault and I have filed it there — but one is a defect in this PR.

## One grounding to fix before merge

    Cysteine-HCl    CHEBI:91247  ->  CHEBI:52891

**`CHEBI:52891` is `QSY9 succinimidyl ester(1+)`** — formula `C43H39N4O13S3`, a fluorescent quencher dye. It is not cysteine in any form.

`CHEBI:91247 L-cysteine hydrochloride` (`C3H7NO2S.HCl`) is correct, and is what MIM holds. The related `L-Cysteine-HCl x H2O` → `CHEBI:17561` (L-cysteine, the free amino acid) also drops the HCl and the water, where MIM has `CHEBI:91248` (the hydrochloride monohydrate).

Both look like the ION-kind rule reaching for a bare amino-acid term and, in the `CHEBI:52891` case, landing on an unrelated compound whose label presumably matched on some substring. Worth checking whether the ION rule produced other ids whose ChEBI **label** is not a recognisable form of the ingredient name — that is the cheap check that catches this class (it is how `KF` → `Lys-Phe` was found in MIM).

## Where this PR is right and MIM is wrong

On the hydrates, **this PR's choices are correct** and MIM is publishing the anhydrous term:

    MgSO4·7H2O     MIM CHEBI:32599 magnesium sulfate    -> #272 CHEBI:31795 heptahydrate  ✓
    CoSO4 x 7 H2O  MIM CHEBI:53470 cobalt(2+) sulfate   -> #272 CHEBI:91244 heptahydrate  ✓
    Na2SeO3·5H2O   MIM CHEBI:48843 disodium selenite    -> #272 CHEBI:131361 pentahydrate ✓
    Na2MoO4·2H2O   MIM CHEBI:75215 (anhydrous)          -> #272 CHEBI:75213 dihydrate     ✓

Filed as MediaIngredientMech#374 (7 records, all with a specific ChEBI hydrate term available) and #375 (the `Na2MoO4·2H2O` case, where MIM's own curation says dihydrate but its label index resolves through a synonym to the anhydrous record).

## Coordination

This PR and **#260** address the same divergence from opposite ends: #272 makes CultureMech internally consistent by picking one id per name; #260 proposes CultureMech resolve from MIM's published `label_index.csv`, which removes competing ids by construction. They are complementary, but **whichever lands first sets the answer for these 27 names**, so it is worth agreeing the direction rather than having them settle it by merge order.

Note also **#265**: CultureMech has 22 distinct synonym strings corpus-wide across 11 of 3,928 names, so it cannot currently resolve name variants on its own — which is the argument for #260's direction.

Issue #279, opened 2026-08-17
Selenate read as sulfate, silicate as sulfite: Se/Si groundings collapsing onto their sulfur analogues


realmarcin:
Found by the element-mismatch detector added for #276/#278.

Three groundings replace selenium or silicon with **sulfur** — the same substitution three
times, which makes it a systematic cause rather than three typos:

| ingredient | grounded to | which is | should be |
|---|---|---|---|
| `H2SeO3` | CHEBI:48854 | **sulfurous** acid | selenous acid (CHEBI:26642) |
| `Na2SeO4 x 10 H2O` | CHEBI:32586 | sodium **sulfate** decahydrate | sodium selenate decahydrate |
| `Na2SiO3 x 5 H2O` | CHEBI:86477 | sodium **sulfite** | sodium metasilicate pentahydrate |

The `H2SeO3` case is fixed (2 rows). **The other two cannot be**: ChEBI has no term for
`sodium selenate decahydrate` or `sodium metasilicate pentahydrate`. Regrounding them to a
near-miss would repeat the error in a new place.

They are recorded in `tests/test_name_term_elements.py` as `KNOWN_WRONG_PENDING_279`,
deliberately separate from the `KNOWN_NO_SALT_TERM` allowlist above it. That distinction
matters: one set is "ChEBI lacks the salt, the free acid is the best available answer", the
other is "this is wrong and we have not fixed it". A test asserts the wrong-set stays at
exactly 2, so it cannot quietly absorb new defects.

Options:
1. **Unground them** — set no `term`, as was done for `Infusion from Potatoes` in #266.
   Honestly unmapped beats confidently wrong, and this is what I would do absent a better
   target.
2. **Ground to the parent** — `selenate` / `silicate` without the hydrate and cation. Less
   wrong, still imprecise.
3. **Request the terms from ChEBI.** Correct in the long run and slow.

Worth chasing the shared cause too. Three independent Se/Si → S substitutions suggests
something upstream — a fuzzy name match scoring `selenate`/`silicate` against `sulfate`, or
a lookup keyed on a normalized string where the distinguishing letters were already lost
(compare the subscript-stripping in #276, where `Na WO .2H O` lost its digits). Whatever
produced these probably produced others that happen to land on a plausible compound and so
are invisible to this detector.

realmarcin:
Fixed in #316 (`0aebc10`), merged.

[...]

Issue #278, opened 2026-08-16
#275 removes the accidental detector for wrong groundings; replace it with a real one


realmarcin:
Found reviewing #275, and caused by it.

Until now, a wrong grounding was visible because `term.label` held the ingredient string
while the id pointed somewhere else — `Na2-EDTA x 2 H2O` sitting next to a term that is
actually *EDTA disodium salt (anhydrous)*. That mismatch is what surfaced the 413 rows
#275 corrected, and before that it is what made #256/#257 findable at all.

#275 sets every label from the ontology. **Every label now agrees with its id, including
any id that is wrong.** The detector was accidental, and it is now gone: after this PR a
mis-grounded row looks exactly like a correct one.

That is still the right change — a label field should hold the label — but it leaves a
hole that should be filled deliberately rather than by accident.

## The replacement

Compare the ingredient's **`preferred_term`** against its grounded term's label, and flag
where they contradict on something checkable. Hydration is the tractable case and was
worth 413 rows in #275 alone:

```
preferred_term: Na2-EDTA x 2 H2O   →  2 waters
term.label:     EDTA disodium salt (anhydrous)  →  0 waters
                                       ^ contradiction
```

The parsing already exists in `audit_grounding_consistency.py` (the `HYDRATE` regex and
its water-count logic) and in the one-off scan used for #275. It needs lifting into a
standing check.

Scope, in order of confidence:

1. **Hydration mismatch** — name says N waters, term says M. Unambiguous. Currently 18
   known survivors (`FeSO4 x 6 H2O` / `x 5 H2O`, `VOSO4 x 5 H2O`) which CHEBI has no term
   for, so the check needs an allowlist or they will be permanent noise.
2. **Salt vs ion** — name says `KNO3`, term label says `nitrate`. Detectable via the
   ION heuristic already in the audit script.
3. **Element mismatch** — name contains `Mg`, term label says `potassium aluminium`.
   Cheap and would have caught the CHEBI:86463 case in #257 directly.

(3) is the one I would add first — it is the crudest and it catches the worst class, the
one where the grounded term is a completely different substance.

Run it as a CI gate with a baseline count, like the other anti-vacuous gates in this repo,
so it cannot silently pass when the detector breaks (see #277 and the classifier bug in
#272 — both were failures that looked like success).

realmarcin:
[...]

Issue #277, opened 2026-08-16
Large PRs silently skip most CI gates: path filters see only the first 300 changed files


realmarcin:
Found on #275. **Four of the six gates did not run, and nothing said so.** The PR shows two
green checks and looks fully verified.

## What happened

GitHub evaluates `on.pull_request.paths` filters against only the **first 300 changed
files** of a PR. #275 changes **20,634** files, and of the first 300, **zero** are under
`data/normalized_yaml/` — they are all `data/merge_yaml/merged/`, which sorts first.

So every workflow filtered solely on `data/normalized_yaml/**` failed to match and never
ran:

| workflow | filters on `merge_yaml`? | ran on #275? |
|---|---|---|
| `label-correspondence` | **yes** | ✅ |
| `tests` (pytest) | no path filter | ✅ |
| `chebi-consistency` | no (only in a comment) | ❌ |
| `concentration-plausibility` | no | ❌ |
| `curation-history` | no | ❌ |
| `validate` | no | ❌ |

`label-correspondence` ran *only* because it happens to also list `data/merge_yaml/**`.

## Why this is worse than a flaky gate

A skipped check is **not** reported as skipped — it is simply absent from `gh pr checks`,
and the PR reads as green. The failure mode is exactly the one this repo keeps hitting: an
output indistinguishable from success. I nearly merged #275 on two green checks.

It also scales the wrong way. The bigger and more sweeping the change — precisely when the
gates matter most — the more likely the first 300 files miss the filter.

## Mitigations, roughly in order

1. **Add `data/merge_yaml/**` to every workflow that filters on `data/normalized_yaml/**`.**
   Cheapest fix, and correct on its own terms: `chebi-consistency`'s own comment already
   says merge_yaml is unguarded and asks for this if it is ever hand-edited. #266 and #275
   both edited it directly, so that condition is now met.
2. **Drop the path filters on the data gates entirely.** They exist to save CI minutes on
   docs-only PRs; the gates are cheap (2–4 min) relative to the risk.
3. **A required "all gates reported" check** that fails when an expected workflow is absent,
   so silence is loud.

(1) plus (3) is what I would do — (1) fixes the immediate hole, (3) stops the next variant
of it.

## For #275 specifically

I ran the four skipped gates locally: `just check-chebi-grounding` exit 0,
`just audit-concentration-plausibility` exit 0, `just validate-strict` 0 errors over
15,877 files. `curation-history` is advisory only and does not block. So that PR is
[...]

Issue #276, opened 2026-08-16
Ingredient groundings include a wrong compound, dropped stereochemistry, and subscript-stripped labels


realmarcin:
Found while working the #260 direction from the MIM side. MediaIngredientMech needs records for the 44 CultureMech ingredient names it lacks, so I verified CultureMech's `term.id` for each before adopting it. **They should not be adopted as-is.**

## 1. One grounding names an unrelated compound

    Sodium sulfide  ->  CHEBI:85357

`CHEBI:85357` is **`3-[(1E,4R)-4-hydroxycyclohex-2-en-1-ylidene]pyruvic acid`**, formula `C9H10O4` — an organic acid. Sodium sulfide is `Na2S`:

    CHEBI:76208  sodium sulfide (anhydrous)   2Na.S
    CHEBI:76209  sodium sulfide nonahydrate   9H2O.2Na.S

This is the same shape as the `Cysteine-HCl` → `CHEBI:52891` (a fluorescent quencher dye) error I flagged on #272. Two independent instances suggests a systematic cause rather than two typos — worth finding whatever produced both.

**The cheap detector**: compare the ChEBI *label* of each assigned id against the ingredient name and flag any pair sharing no token. It is how `KF` → `Lys-Phe` was caught in MIM.

## 2. Stereochemistry silently dropped

    DL-Tyrosine    ->  CHEBI:17895  L-tyrosine
    DL-threonine   ->  CHEBI:16857  L-threonine

A racemate is not its L-enantiomer. `CHEBI:18186 tyrosine` is stereo-unspecified and closer; ChEBI has no `DL-threonine` term, so that one may need a registry mint rather than a wrong answer.

## 3. Labels have lost their subscripts

Eight ingredient names are unreadable as chemistry because the digits are gone:

| stored label | almost certainly |
|---|---|
| `H BO` | H₃BO₃ boric acid |
| `ZnCl .6H O` | ZnCl₂·6H₂O |
| `Co(NO ) .6H O` | Co(NO₃)₂·6H₂O |
| `LiCl.H O` | LiCl·H₂O |
| `As O` | As₂O₃ |
| `Na WO .2H O` | Na₂WO₄·2H₂O |
| `TeO` | TeO₂ |
| `MgSO .7H2O` | MgSO₄·7H₂O |

[...]

PR #280, opened 2026-08-17
Detect groundings whose term lacks an element the name demands (#276, #278)

realmarcin:
Addresses #278's element-mismatch item and #276's "cheap detector" — they asked for the
same thing.

## Why now

Two groundings named a completely different kind of molecule — `Cysteine-HCl` on a
fluorescent quencher dye, `Sodium sulfide` on an organic acid — and **both were found by
hand**. Two independent instances is a pattern.

It is also newly *necessary*: since #275 refilled `term.label` from the ontology, a wrong
id no longer looks wrong. `preferred_term` is the last independent signal.

## Names don't work; elements do

| check | `MgSO4 x 7 H2O` vs *potassium aluminium sulfate* | `KNO3` vs *potassium nitrate* |
|---|---|---|
| word overlap | **passes** (shares "sulfate") — the exact #257 error | **flags** — a correct grounding |
| elements | flags (no Mg) ✅ | passes ✅ |

The cation is the discriminator. So this compares element sets against ChEBI's own
`chemrof:generalized_empirical_formula`.

## Calibration was the work

The first run reported 75 pairs, mostly noise: `Fe(III) citrate` read `(III)` as iodine,
`Vitamin B12` read `B` as boron, `PABA` read `P` as phosphorus, and every vanadyl sulfate
row was reported as lacking vanadium — the formula parser couldn't see `V` inside the token
`OV`. Down to 10 after fixes, then **5 more appeared** when the scan went line-based, which
(as with #278's hydration check) has strictly better coverage than the yaml walk.

## Found and fixed — 95 rows

Four named a different substance entirely:

| ingredient | was grounded to | now |
|---|---|---|
| `Sodium sulfide nonahydrate` (35) | an organic acid | CHEBI:76209 |
| `Sodium sulfide` (8) | an organic acid | CHEBI:76208 |
| `Manganese sulfate monohydrate` | **ninhydrin** | CHEBI:86364 |
| `Na WO .2H O` | **a peptide** | CHEBI:63939 |
| `Co(NO ) .6H O` | an organic | CHEBI:86214 |
| `H2SeO3` | sulfurous acid | CHEBI:26642 |
| `Iron(II) sulfate` | the bare iron(2+) ion | CHEBI:75832 |

## Recorded, not tolerated — #279

`Na2SeO4 x 10 H2O` is grounded to sodium **sulfate** and `Na2SiO3 x 5 H2O` to sodium
**sulfite** — the same Se/Si → S collapse that put `H2SeO3` on sulfurous acid. ChEBI has no
term for either correct compound, so regrounding would repeat the error elsewhere.

They sit in `KNOWN_WRONG_PENDING_279`, kept **deliberately separate** from the
`KNOWN_NO_SALT_TERM` allowlist. "ChEBI lacks the salt, the free acid is the best answer"
and "this is wrong and unfixed" must not blur together. A test pins the wrong-set at
exactly 2 so it can't quietly absorb new defects.

Three independent Se/Si→S substitutions suggests a shared upstream cause worth chasing —
whatever produced them likely produced others that happen to land on a *plausible* compound
and are therefore invisible to this detector.

## Verification

- `pytest`: 1,206 passed, 4 new
- the standing gate carries an anti-vacuous test built from every real error **and** every
  real false positive above, so a parser that stops parsing fails loudly — the failure mode
  of #277, the #272 classifier, and the `yaml.dump` no-op in #275
- both corpora


realmarcin:

**1. Is the detector's output trustworthy, or tuned until it agreed with me?** This is the
real risk with a heuristic. The calibration set is built from cases decided *independently*
of the detector: `CHEBI:86463` (found in #257 by label mismatch), `Cysteine-HCl` (found by
hand in #270), `Sodium sulfide` (found from the MIM side in #276). The detector has to
catch all three and must not flag `MgSO4 x 7 H2O`, `KNO3`, or `VOSO4 x 2 H2O`. Those
assertions are in the test, so tuning it to pass the corpus cannot quietly break them.

**2. Did the fixes land where intended?** 95 rows across both corpora, residual
element-mismatch pairs down to the 5 allowlisted ones. `test_name_term_elements` passes.

**3. Is the allowlist doing honest work?** It is split in two on purpose —
`KNOWN_NO_SALT_TERM` (ChEBI genuinely lacks the salt; the free acid is the best available
answer) and `KNOWN_WRONG_PENDING_279` (this is wrong and unfixed). Blurring them would let
a real defect hide behind "ChEBI limitation". A test pins the wrong-set at exactly 2.

**4. The CI failure was the #145 drift class**, not the change: `derived_artifacts.tsv` is
computed from `git ls-files`, so a new report is invisible while untracked. It passed
locally *before* the commit and failed *after* — the reverse of the usual order, which is
why it is easy to miss. Regenerated with the report tracked.

## Filed

- #279 — `Na2SeO4`→sodium **sulfate**, `Na2SiO3`→sodium **sulfite**, the same Se/Si→S
  collapse as `H2SeO3`→sulfurous acid. Unfixable: ChEBI has no term for either correct
  compound. Three independent instances of one substitution points at a shared upstream
  cause, and anything it produced that landed on a *plausible* compound is invisible to
  this detector.

No blocking findings.

Issue #263, opened 2026-08-15
34 records carry YAML anchors/aliases from the pre-deepcopy split path


realmarcin:
Found reviewing #261, which fixed the cause but not the existing records.

`apply_cocktail_nesting.split_parts` used `dict()` to copy an ingredient being split
across two stocks. That is a **shallow** copy, so the nested `term` and
`mediaingredientmech_chebi_term` dicts stayed shared with the source ingredient — and the
two copies land in *different* `solutions:` entries. `yaml.dump` serializes a shared
object as an anchor plus an alias:

```yaml
  - preferred_term: Nicotinic acid
    term: &id001
      id: CHEBI:15940
      label: Nicotinic acid
  ...
  - preferred_term: Nicotinic acid
    term: *id001
```

Why it matters beyond aesthetics: the two solutions alias **one object**. Any consumer
that loads the record and edits one solution's `term.label` silently edits the other's.
It also makes review harder — a reviewer cannot see the aliased value at the alias site.

#261 switched to `copy.deepcopy`, so no new records get anchors (verified: the count
stayed at 34 across a 94-record apply). The 34 written earlier still have them:

```
$ grep -rl "&id00" data/normalized_yaml | wc -l
34
```

Fix is a load-with-alias-expansion and re-dump over those 34. Worth confirming while
doing it that the aliased values really are identical in each case, rather than assuming
the alias was harmless.

realmarcin:
**Correcting this issue — I filed it with the wrong cause.**

I attributed all 34 anchor-carrying records to the shallow-copy bug in
`apply_cocktail_nesting.split_parts`. That accounts for **17** of them. The other 17 have
a different and more serious problem, and the anchors there are a symptom rather than the
defect.

Breaking the 34 down by shape:

| shape | records | what it is |
|---|---:|---|
| aliased nested `term` dicts, has `solutions:` | 17 | the `split_parts` shallow-copy bug — genuinely just serialization |
| the **same ingredient object twice in `ingredients:`** | 17 | duplicate ingredient rows |

The second class is a data bug, not a formatting one. `for_dsm_14290.yaml` lists:

```
Tryptone        10.0 G_PER_L   (twice)
Yeast extract    5.0 G_PER_L   (twice)
Sodium chloride 10.0 G_PER_L   (twice)
```

Anything that sums or iterates ingredients sees this recipe as double-strength in three
components.

And the duplication is **not** confined to the anchored records — YAML anchors are just
[...]

PR #282, opened 2026-08-17
Fix dropped stereochemistry and restore subscript-stripped names (#276)

realmarcin:
Closes the remaining two items of #276. (Item 1, the wrong compound, was fixed in #280.)

## Item 2 — a racemate is not its L-enantiomer

| ingredient | was | is now | rows |
|---|---|---|---:|
| `DL-Tyrosine` | CHEBI:17895 **L-tyrosine** | CHEBI:18186 tyrosine | 36 |
| `DL-threonine` | CHEBI:16857 **L-threonine** | CHEBI:26986 threonine | 2 |

Smaller than the issue implied, and worth saying why rather than quietly doing less: **most
`DL-` names in the corpus are already correct.** `DL-methionine`→*methionine*,
`DL-Malic acid`→*malic acid*, `DL-Calcium pantothenate`→*Calcium pantothenate* are all
stereo-**unspecified** terms and fine as they stand. Only two named a single enantiomer.

## Item 3 — digits restored, 51 names across both corpora

| corrupted | repaired | | corrupted | repaired |
|---|---|---|---|---|
| `H BO` | `H3BO3` | | `As O` | `As2O3` |
| `Co(NO ) .6H O` | `Co(NO3)2.6H2O` | | `K SO` | `K2SO4` |
| `Na WO .2H O` | `Na2WO4.2H2O` | | `K CO` | `K2CO3` |
| `MnSO .H O` | `MnSO4.H2O` | | `LiCl.H O` | `LiCl.H2O` |
| `Ca(NO ) .4H O` | `Ca(NO3)2.4H2O` | | `Na SO .10H O` | `Na2SO4.10H2O` |
| `Fe(NH ) (SO ) .6H O` | `Fe(NH4)2(SO4)2.6H2O` | | | |

Each is listed **explicitly, not inferred by rule**, and corroborated by its grounded term
where it has one — `As O` sits on *diarsenic trioxide*, which is As₂O₃.

**`Na VO` is deliberately not repaired.** Na₃VO₄ (ortho) and NaVO₃ (meta) are both
plausible and the corpus cannot disambiguate. Guessing there would be the same class of
error this repair exists to undo. A test asserts it stays out.

The corrupted original is preserved in each ingredient's `notes` — the source string is
otherwise unrecoverable, and a reader should see that the name was repaired rather than
transcribed.

## The baselines go UP, and that's the finding

Implausible rows **9,750 → 9,757**, cocktails **185 → 186**. The gate says in terms *"Do
NOT raise the baseline to make this pass"* — so I checked before touching it.

**The corruption was hiding rows from the detectors.** `H BO` matches no trace-element
pattern and was flagged **0 times** on main; `H3BO3` matches and is flagged. Those
concentrations were always wrong — the broken name made them invisible. Verified by
counting both spellings on either side of the change rather than assuming.

So repairing names surfaced 7 pre-existing implausible rows and 1 pre-existing flattened
cocktail. Both baselines carry that explanation inline so the next reader doesn't
mistake it for a cover-up.

## Note on the encoding bug itself

`MgSO .7H2O` keeps the `2` in `H2O` but loses the `4` — ASCII digits survived while
subscript characters (₂₃₄) did not. That points at an encoding step upstream, not a parser,
and **it is still unfixed**; this only repairs the damage it left. Worth chasing separately.

## Verification

- `pytest`: 1,207 passed, 5 new
- both corpora, so the #266 derived-corpus check stays green
- canary first, before fan-out


realmarcin:

**1. Did renaming collide with an existing row?** The sharp risk: renaming `As O` →
`As2O3` inside a record that already holds `As2O3` recreates the duplicate class #263 just
removed. Checked corpus-wide — 4 collisions, all in `ucm.yaml`.

**They are not caused by this PR.** `ucm.yaml` carries *two* `As O` rows and two
`Na WO .2H O` rows on main already; the repair renamed both consistently. Verified against
`origin/main` rather than assumed. Filed as #283, because the pair differs by exactly
**1000×** (0.000093 vs 0.093 g/L) — the flattened-cocktail signature from #150, with stock
strength and final concentration sitting side by side.

**2. Do the standing gates still hold?** `test_corpus_grounding_agreement` and
`test_grounding_hydration_agreement` both pass — renaming keys the cross-corpus check, so
running both corpora mattered.

**3. Were the baselines raised honestly?** This is the part that deserves scrutiny, since
the gate says in terms *"Do NOT raise the baseline to make this pass"*. `H BO` was flagged
**0 times** on main and `H3BO3` is flagged — counted on both sides of the change, not
inferred. The corruption was hiding rows from the detector; the concentrations were always
wrong. Both baselines carry that reasoning inline.

**4. Was anything lost?** The corrupted original is preserved in each ingredient's `notes`,
and existing notes are appended to rather than replaced — a test covers that.

## Filed

- #283 — `ucm.yaml`'s 1000×-apart duplicate rows, and the observation that
  `audit_concentration_plausibility` checks absolute magnitude but never the *ratio between
  sibling rows*, so this shape has no detector at all.

## Worth stating

`Na VO` is deliberately unrepaired (Na₃VO₄ vs NaVO₃ — the corpus cannot say which), and the
upstream encoding bug that strips subscripts is **still unfixed**; this only repairs its
output. It will keep producing these.

PR #285, opened 2026-08-17
Upgrade deep-research-client 0.2.4 → 0.2.10 and record what it actually offers (#284)

### LINKED ISSUE #284: deep-research-client is 6 releases stale (0.2.4 vs 0.2.10); the reason we bypass it no longer holds
We do use it — as a **CLI subprocess**, not a Python import. `scripts/research_media.py`
shells out to `deep-research-client` (`--client-command`, default `deep-research-client`),
and it is declared in the `dev` extra as `deep-research-client[cyberian]>=0.2.4`. Nothing
imports `deep_research_client` in Python. It resolves from **PyPI**, not from the DisMech
repo.

It is installed and the CLI works. The problem is that it is stale, and staleness has
cost us work.

## Version state

| | version | providers available |
|---|---|---|
| CultureMech (installed) | **0.2.4** | `cyberian` only — verified with `deep-research-client providers` |
| PyPI latest | **0.2.10** | OpenAI, **Edison**, Asta, Perplexity, Consensus, Cyberian, OpenScientist, **Claude Code (local CLI)** |
| DisMech | **0.1.3**, declared unpinned | older still |

## Two consequences

**1. The documented reason for bypassing it is obsolete.** `scripts/research_media_edison.py`
says:

> Uses the `edison-client` SDK directly. The companion `research_media.py` wraps
> `deep-research-client`, but **as of DRC 0.2.4 only `cyberian` and `openai` are registered
> providers — Edison/PaperQA is not exposed there**, so we drive the SDK directly.

0.2.10 exposes Edison (`falcon`/`edison`/`eds`/`science`, "powered by PaperQA3"). So the
whole second code path exists to work around a limitation that has since been lifted, and
we now maintain two research entry points instead of one.

**2. We reimplemented a provider the library already has.** When Edison started returning
HTTP 402 earlier, the fallback was a hand-rolled native-Claude research path in
`.claude/skills/deep-research-medium/SKILL.md`. 0.2.10 ships **Claude Code (local CLI)** as
a first-class provider behind the same interface. That fallback could have been a
`--provider` flag.

## Also worth checking on upgrade

Our `.env.example` sets `EDISON_PLATFORM_API_KEY` (with a note that the SDK reads it
natively), but DRC's own table expects **`EDISON_API_KEY`**. If the Edison provider is
adopted through DRC rather than the SDK, that name has to be reconciled or the provider
will silently not auto-detect — DRC "auto-detects providers from environment variables",
so a wrong name presents as "provider unavailable", not as an error.

## Suggested

1. Bump to `>=0.2.10`, re-run `deep-research-client providers`, and record what actually
   registers — the `[cyberian]` extra may need widening for the others.
2. If Edison registers, retire `research_media_edison.py`'s SDK bypass and its docstring
   rationale, collapsing to one entry point.
3. Evaluate the Claude Code provider against the hand-rolled skill path; keep whichever is
   better, but not both by accident.
4. Coordinate with DisMech, which is on 0.1.3 and unpinned — an unpinned shared dependency
   across repos is how the two drift apart silently.

Not urgent in the sense that nothing is broken today; the research lane is dormant. Urgent
in the sense that it is the lane we reach for when the corpus needs evidence (#150's 185
cocktails, #279, #273 all need source reading), and it should not be discovered stale at
the moment of need.

realmarcin:
Closes #284. Providers **verified**, not assumed.

## Before / after

`deep-research-client providers`, at 0.2.4:

```
cyberian
```

At 0.2.10, with `.env` loaded:

```
falcon        ← Edison Scientific, via EDISON_API_KEY
cyberian
claude_code   ← available with NO credentials
```

Both #284 claims confirmed. Edison **is** now reachable through DRC, and the Claude Code
provider **does** exist — so the hand-rolled native-Claude fallback in
`.claude/skills/deep-research-medium/SKILL.md`, written when Edison started returning 402,
duplicates something the library ships.

## #284's main recommendation was wrong, and checking is what showed it

I proposed retiring `research_media_edison.py`'s SDK bypass and collapsing to one entry
point. **Don't.** `providers --provider falcon --show-params` exposes only:

`system_prompt`, `allowed_domains`, `temperature`, `max_tokens`, `max_embedded_images`

It does **not** expose Edison's **job** selection — and job choice is the entire point of
that script. `LITERATURE` vs `LITERATURE_HIGH` vs `precedent` vs `phoenix` are different
agents at different costs. Routing through DRC today would silently pin us to one job.

So both paths stay. What changes is that the second one now states its **current** reason
instead of an expired one — its docstring still said *"as of DRC 0.2.4 only cyberian and
openai are registered providers"*, which stopped being true six releases ago.

## The real trap was in `.env.example`

The two consumers read **different variables**:

| consumer | reads |
|---|---|
| `research_media_edison.py` (edison-client SDK) | `EDISON_PLATFORM_API_KEY` (alias: `EDISON_API_KEY`) |
| `deep-research-client` | **`EDISON_API_KEY` only** |

DRC auto-detects providers from the environment, so the wrong name **does not raise** —
Edison simply never appears in the provider list, which reads as *"not configured"* rather
than *"misconfigured"*. `.env.example` documented only the SDK name, so anyone following it
would have had a DRC install that silently could not see Edison.

It now documents both and recommends `EDISON_API_KEY`, which satisfies both.

The live `.env` already used `EDISON_API_KEY`, so nothing was actually broken — which is
precisely why it could have sat there indefinitely. Same silent-failure shape as #277's
absent CI gates.

## Guard

The test asserts the pin **floor**, not the installed version (CI resolves from the lock; a
developer may be ahead), that the **lockfile satisfies that floor**, and that the SDK
script still explains why it exists — if someone deletes that justification, the script
should go with it rather than linger unexplained.

Per instruction, **nothing changed in the dismech repo**. It remains unpinned on 0.1.3 and
will drift independently.

## Verification

- `pytest`: 1,214 passed, 3 new; `tests/test_research_media.py` (13) still green
- providers listed before and after, with and without `.env`
- no API calls made — provider registration only

### DIFF
--- a/.claude/skills/deep-research-medium/SKILL.md
+++ b/.claude/skills/deep-research-medium/SKILL.md
@@ -1,17 +1,43 @@
 ---
 name: deep-research-medium
-description: Run Edison Scientific deep research for a CultureMech medium record, then for each candidate organism returned by phase 1 run a per-organism follow-up that extracts the recipe, culture conditions, identifiers, and citation metadata from the primary publication. Produces a curation-focused report with proposed YAML changes.
+description: Run deep research for a CultureMech medium record through whichever provider fits the question (claude_code, openscientist, codex), then for each candidate organism returned by phase 1 run a per-organism follow-up that extracts the recipe, culture conditions, identifiers, and citation metadata from the primary publication. Produces a curation-focused report with proposed YAML changes. Start at 'Choosing a provider' — Edison/falcon is billing-blocked.
 category: research
 requires_database: false
 requires_internet: true
-version: 1.0.0
+version: 2.0.0
 ---
 
-# Deep Research for a Medium (Edison API)
+# Deep Research for a Medium
+
+## Choosing a provider — read this first
+
+Measured on one shared question, at deep-research-client 0.2.10 (#284):
+
+| provider | time | chars | URLs | PMIDs | reach for it when |
+|---|---:|---:|---:|---:|---|
+| `claude_code` | 295s | 20,876 | **22** | 5 | you need MANY citable sources. No credentials. **Default.** |
+| `openscientist` | 490s | 21,518 | 6 | **2** | reasoning depth matters more than breadth; PMID-anchored |
+| `codex` | 439s | 10,813 | **27** | – | cheap sweeps over many records (`just research-media-codex`) |
+| `falcon` (Edison) | 6s | – | – | – | **DO NOT** — returns HTTP 402 Payment Required |
+
+For corpus work where a grounding has to be *settled* — #150's cocktails, #279's
+compounds — prefer `claude_code`: breadth of citable sources is the deliverable.
+
+**Two traps, both of which have cost time already:**
+
+1. **"Available" ≠ working.** `deep-research-client providers` lists a provider once its
+   env var is set, and says nothing about whether the account can run a job. `falcon`
+   lists as available and 402s on every call.
+2. **Depth is prompt-bound, not tool-bound.** A bare one-line question made `codex` look
+   shallow (3 sources); the same tool through the real template returned 27. Always run
+   through the filled template, never an ad-hoc question.
+
+These numbers are one question in one domain. Treat them as measured guidance, not law —
+if a second data point disagrees, update the table rather than defending it.
 
 ## Overview
 
-**Purpose**: For one CultureMech medium record, run a two-phase Edison
+**Purpose**: For one CultureMech medium record, run a two-phase
 Scientific deep-research workflow:
 
 1. **Phase 1 — medium-level literature search.** Submit the medium's
@@ -50,11 +76,18 @@ formulation** of one medium and **each of its modeled variants** against
 authoritative sources, returning curator-ready corrections rather than new
 organisms.
 
-### Native Claude Code path (default — no Edison credits)
+### Default path — `claude_code` provider (no credentials, no credits)
 
-Run the validation with this harness's own web tools (or the `deep-research`
-skill) instead of the Edison API. Render the filled prompt, then research
-it yourself:
+This used to be a hand-rolled instruction to do the searching yourself, written
+when Edison started returning 402. That is now redundant: `deep-research-client`
+ships `claude_code` as a first-class provider, so prefer
+
+```bash
+just research-media claude_code <SLUG-OR-PATH>
+```
+
+Render-only remains useful when you want to inspect or hand-edit the brief before
+spending time on it:
 
 ```bash
 # print the validation prompt for one medium + its variant set:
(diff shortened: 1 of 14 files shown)


realmarcin:
Rebased onto current `main` (six merges landed while this was open) and reviewed adversarially. **Two defects fixed here; one filed.**

## Rebase

One conflict, in the `DEEP RESEARCH` comment block of `project.justfile`: `main` had gained the entity-runner contract note (#289) where this branch adds the measured provider table. Both are worth keeping, so both are kept — contract first (what the recipe *is*), then the table (which provider to pick). Resolved so the later cyberian commit still applied cleanly on top, which it did.

## Fixed — the PR contradicted itself

The justfile table this branch adds records:

```
falcon    6s   -  HTTP 402 Payment Required
cyberian 21s   -  HTTP 500 - agentapi not running
```

…while the triage tool it ships routed **every stage** to falcon:

```
Route now: falcon (scientific evidence synthesis); cross-check/fallback: claude_code
```

because "available" only ever meant *an env var is set*. `KNOWN_BLOCKED` now records what actually happened when each provider was called, so a measured-dead provider reports `blocked` however well its credential is configured. Routes to `claude_code` now.

## Fixed — machine-readable output was self-contradictory

`--provider X --json` filtered `ranking` but left the recommendation untouched:

```
ranking=['asta']   recommended=claude_code   fallback=cyberian
```

Only machine consumers saw it — the human path takes a different branch. One `recommendable()` now serves both paths, and the filtered document recomputes from what survived.

Both are #290, which I filed earlier against this code; it can close with this PR.

## Added — the policy knobs #290 asked for

`--allow a,b` and `--no-paid`. `medium` is deliberately **not** "paid": `claude_code` is the medium-cost provider and keeping it is usually the point of asking. An unknown name in `--allow` is rejected rather than silently matching nothing, and the allowlist bounds the *recommendation* while leaving the full ranking visible.

## One subtlety worth flagging to a reviewer

Blocking short-circuits the credential check, so `provider_status` was split from `credential_status`. Without that split, adding a provider to `KNOWN_BLOCKED` would silently drop the test that its env-var aliases are spelled right — which is exactly what `test_falcon_platform_key_is_recognized_without_exposing_it` covers. It now asks the lower-level question and still passes.

## Filed, not fixed — #311

`research_media_codex.py` sits outside the entity-runner contract: no `--focus`, no provider argument, its own output scheme. The justfile table lists codex as a peer of `claude_code` and `openscientist`, which is a reading the interface doesn't support. Not blocking — the script works (verified by dry run) — but the interface should converge.

## Verification

- `just research-media-codex ko2_no3 --dry-run` renders correctly.
- `--no-paid` and `--allow asta` both route as intended; a bad `--allow` exits 1.
- **1,331 tests pass**, ruff clean.
- No provider called, no credits — dry-run and env-var fixtures throughout.

PR #291, opened 2026-08-18
Report the sidecars this Edison run wrote, not the ones lying around (#288)

### LINKED ISSUE #288: Propagate current-run Edison sidecar provenance tracking across Mechs
## Problem

TraitMech contains a correctness fix in scripts/_edison_capture.py that records which sidecars the current invocation actually wrote.

CultureMech, MediaIngredientMech, and CommunityMech still determine sidecar presence from files already on disk. Edison output stems are deterministic. On a rerun, a prior response, agent-state, or files sidecar can remain present even when the new invocation did not fetch it, causing the new metadata record to attribute stale provenance to a different task_id.

This is an auditability defect in shared paid-research capture code.

## Proposed scope

1. Treat the TraitMech current-invocation written-set behavior as the candidate canonical fix.
2. Add a regression test with pre-existing sidecars and a new response that does not rewrite all of them.
3. Propagate the fixed capture module and tests to CultureMech, MediaIngredientMech, and CommunityMech.
4. Put _edison_capture.py under shared drift enforcement or move it to a shared package.
5. Confirm enrich_edison_response and dry-run behavior remain correct.

## Acceptance criteria

- Metadata reports true only for sidecars produced by the current invocation.
- A stale sidecar from a previous run is never attributed to the new task_id.
- All four Edison-capable Mechs run the same regression test.
- Shared capture code is protected from future unreviewed drift.
- Tests use fixtures or mocks and spend no provider credits.

Fleet review: CultureBotAI/culturebotai-claw docs/reviews/five_mech_shared_functionality_review.md

realmarcin:
Closes #288 for CultureMech.

## The defect

Edison output stems are deterministic — `{slug}-edison-{job}` — so re-running
the same medium and job writes into a directory that already holds the previous
run's sidecars. `capture_full_response` built its `sidecar_files` provenance
block from a plain `.exists()` sweep of that directory:

```python
"sidecar_files": _existing_sidecars(out_dir, stem),
```

A rerun whose verbose fetch or `list_files` call failed therefore wrote
`agent_state_json: true` into a meta stamped with a **new** `task_id`, pointing
at a trace belonging to the old one. For a field whose entire purpose is
provenance, an auditor following the meta reads the wrong trajectory.

## The fix

Track the keys the invocation actually writes, and report the AND of "we wrote
it" and "it is on disk". This is TraitMech's fix, ported.

The stale files stay where they are — only the claim about them changes.
Deleting them would destroy the earlier run's evidence.

`_existing_sidecars` keeps its disk-snapshot behaviour when the written-set is
omitted, which is what `enrich_edison_response` needs: backfill runs against the
`task_id` already stored in the meta, so there "what is on disk for this stem"
*is* the truthful answer. That call site is unchanged, and a test now pins why
it is allowed to differ.

## Tests

TraitMech shipped this fix with no test, so all six cases are new:

| test | pins |
|---|---|
| `test_stale_sidecars_are_not_attributed_to_the_new_task` | the regression itself — prior sidecars from `task-OLD`, new response, neither refetched |
| `test_sidecars_this_run_did_write_are_reported_true` | the fix does not just report `False` for everything |
| `test_empty_output_dir_reports_only_what_was_written` | first run of a stem |
| `test_written_set_cannot_claim_a_file_that_is_absent` | a failed write never yields a path that is not there |
| `test_omitting_the_written_set_still_snapshots_the_disk` | the `enrich_edison_response` contract |
| `test_dry_run_reports_no_sidecars_at_all` | a dry run inherits nothing and overwrites nothing |

Verified the regression tests bite — reverting only the source change:

```
FAILED test_stale_sidecars_are_not_attributed_to_the_new_task
FAILED test_written_set_cannot_claim_a_file_that_is_absent
2 failed, 4 passed
```

**No credits spent.** No Edison client is constructed: the response is a stub
object and `client=None` exercises the skipped-fetch path. 127 edison/research
tests pass.

## Scope note

Issue #288 also asks to propagate the fix to MediaIngredientMech and
CommunityMech and put `_edison_capture.py` under shared drift enforcement. Those
are spoke-side changes in other repositories and are deliberately not in this PR;
the new docstring says "record" rather than "medium" so this copy stays adoptable
verbatim. Follow-up issues will be filed in those repos with CultureMech as the
canonical source.

The two ruff findings in `_edison_capture.py` (`I001`, `B009`) predate this
change — confirmed identical on `origin/main` — and are left alone.

### DIFF
--- a/scripts/_edison_capture.py
+++ b/scripts/_edison_capture.py
@@ -392,14 +392,19 @@ def capture_full_response(
     answer_reasoning = _get_attr(response, "answer_reasoning")
     body = formatted_answer or answer or "(no answer field on this job's response type)"
     md_path.write_text(body)
+    # Track what THIS invocation wrote, so the meta reports its own provenance
+    # rather than whatever a previous run of the same stem left on disk.
+    written: set[str] = {"answer_md"}
 
     # Full raw response dump (every SDK field, future-proof)
     response_dump = _safe_model_dump(response)
     response_json_path.write_text(json.dumps(response_dump, indent=2, default=str))
+    written.add("response_json")
 
     # Citations sidecar
     citations = parse_citations(formatted_answer or answer)
     citations_md_path.write_text(render_citations_md(citations, query=query))
+    written.add("citations_md")
 
     # Verbose + files via secondary fetches (no extra billing)
     task_id = str(_get_attr(response, "task_id") or "")
@@ -415,12 +420,14 @@ def capture_full_response(
                 "environment_frame": environment_frame,
                 "metadata": verbose_metadata,
             }, indent=2, default=str))
+            written.add("agent_state_json")
 
     files_listing = _try_list_files(client, task_id) if task_id else None
     artifacts_manifest: list[dict[str, Any]] = []
     if files_listing is not None:
         files_path.write_text(json.dumps(_safe_model_dump(files_listing),
                                          indent=2, default=str))
+        written.add("files_json")
         # Pull the actual content of named curation artifacts (small,
         # not internal PaperQA pickles) into a sibling artifacts/ dir.
         artifacts_manifest = fetch_named_artifacts(
@@ -453,7 +460,7 @@ def capture_full_response(
         "answer_reasoning_chars": len(answer_reasoning or ""),
         "citations_parsed": len(citations),
         "query_sha256": query_sha256(query),
-        "sidecar_files": _existing_sidecars(out_dir, stem),
+        "sidecar_files": _existing_sidecars(out_dir, stem, written),
         "artifacts_fetched": [a for a in artifacts_manifest if a.get("status") == "fetched"],
         "artifacts_skipped": [a for a in artifacts_manifest if a.get("status") != "fetched"],
     })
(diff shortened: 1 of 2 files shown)


realmarcin:
Adversarial review pass (read-only) — two things checked beyond the diff.

**1. Is the committed corpus already corrupted?** No. Audited all 131 `research/**/*-meta.yaml` by comparing each meta's `task_id` against the `task_id` recorded *inside* the agent-state trace it claims:

```
metas claiming agent_state_json=true, id MATCHES trace : 102
metas claiming agent_state_json=true, id MISMATCHES    : 0
metas claiming agent_state_json=true, FILE ABSENT      : 0
metas not claiming agent_state_json                    : 29
```

So the bug was real but had not yet bitten, and this PR needs no corpus back-fix. Worth stating explicitly given how often derived artifacts in this repo drift out from under a fix. (`agent-state.json` is the only sidecar carrying an inner `task_id`, so it is the only one auditable this way — `files.json` cannot be checked the same way.)

**2. Can anything undo the fix?** Yes — filed as #292. `enrich_edison_response.needs_enrichment` decides what to refetch from bare disk existence, so a stale trace is neither noticed nor refetched, and the meta update then re-stamps a `.exists()` snapshot. In a narrow window (stale sidecar present **and** at least one other sidecar missing, so the `already-complete` early return does not fire) that flips a corrected `false` back to `true`.

Not folded into this PR: enrichment legitimately reports files it did not write this run, so it needs different semantics and its own tests. The docstring here claims the disk snapshot is truthful for `enrich_edison_response` because it runs against the `task_id` already in the meta — true of the meta, not guaranteed of the file. #292 is that gap.

realmarcin:
Second adversarial pass. No defects found; recording what was checked so the merge is not on faith.

**The semantic change is safe for every consumer.** The concern was that some caller treats `sidecar_files` as a disk-presence oracle and would now skip work it should do. Exhaustive grep across the repo:

```
scripts/research_media_edison.py:277           reads -> print only
scripts/research_organism_recipe_edison.py:191 reads -> print only
scripts/_edison_capture.py:456                 writes
scripts/enrich_edison_response.py:203          writes
```

No control flow anywhere depends on it. `enrich_edison_response.needs_enrichment` decides what to refetch from `.exists()` directly, not from the meta, so it is unaffected by this change (its own separate hole is #292).

**The written-set is added at exactly the write sites.** `agent_state_json` is added inside the same `if any(...)` guard as its write, and `files_json` inside the same `if files_listing is not None` guard — so there is no path that writes a file without recording it, or records one it did not write. `answer_md`, `response_json` and `citations_md` are unconditional, matching their unconditional writes.

**The reporting is an AND, not a substitution.** `{key: (key in written and value)}` means a key in the written set whose file is absent still reports `False`, so the meta can never point at a path that does not exist. Covered by `test_written_set_cannot_claim_a_file_that_is_absent`.

**Corpus re-verified**: all 131 committed `research/**/*-meta.yaml` still show 102 claiming `agent_state_json: true` with 0 `task_id` mismatches against the trace they name — no back-fix needed.

CI green (pytest). Note only the pytest gate ran here: the other workflows have path filters and this diff touches only `scripts/` and `tests/`, which is correct behaviour, not a skipped check.

Issue #289, opened 2026-08-18
Standardize deep-research citation artifacts and entity-runner contract across Mechs


realmarcin:
## Problem

The five generic deep-research runners do not expose one consistent operational contract.

Citation handling differs: TraitMech intentionally disables deep-research-client separate citation sidecars because the regex output was malformed and duplicative, while CultureMech, MediaIngredientMech, CommunityMech, and ProteinTraitsMech still request them.

Target interfaces also differ:

- CultureMech, CommunityMech, and ProteinTraitsMech accept provider plus a flexible target.
- TraitMech requires category and slug, with provider passed as an option.
- MediaIngredientMech requires provider, status, and slug.

Secondary research focuses also rank providers but remain disconnected from runner template selection.

## Proposed scope

1. Define a fleet-level research artifact contract: report, citations, metadata, cache identity, and failure semantics.
2. Decide whether separate citation files are authoritative, derived, or disabled until the upstream parser is fixed.
3. Add a common research-entity provider target focus interface, implemented by thin Mech-specific resolvers.
4. Map each configured focus to the appropriate template rather than always using the default template.
5. Retain compatibility aliases for existing just recipes.
6. Add contract tests shared across all five repositories.

## Acceptance criteria

- The same conceptual command works across all five Mechs.
- Every configured focus selects an appropriate template and output label.
- Citation artifacts have one documented, tested meaning.
- Existing malformed sidecars are not treated as authoritative evidence.
- Output paths, provider aliases, dry-run behavior, and failures are consistent.
- Contract tests invoke no research provider and spend no credits.

Fleet review: CultureBotAI/culturebotai-claw docs/reviews/five_mech_shared_functionality_review.md

realmarcin:
CultureMech's part is done in #293. Also correcting one premise.

**"The five generic deep-research runners" is four.** proteintraitsmech has no deep-research runner on GitHub at all — `scripts/` contains no `research_*.py`. The runners are `research_media.py` (CultureMech), `research_trait.py` (TraitMech), `research_ingredient.py` (MediaIngredientMech), `research_community.py` (CommunityMech). Same root cause as #287: the review counted files that exist only as untracked working-tree copies in local clones.

[...]

PR #293, opened 2026-08-18
Give the media runner one artifact contract and a focus that actually picks the prompt (#289)

realmarcin:
Addresses #289 for CultureMech. Scope is this repo only — see the note at the end.

## 1. Citation artifacts now have one meaning

`research_media.py` asked deep-research-client for a `--separate-citations`
sidecar. That file is a **regex over the report prose**, not structured output
from the provider. CultureMech produced exactly one before this change,
`research/media/algae/2asw-deep-research-falcon.md.citations.md`:

- it re-emits the entire ~55-line rendered prompt as "Query" — content the
  report already carries as template variables;
- entry **12 of 27** is the bare string `Na+`;
- `10.1101/2024.06.09.598106` appears **three times**, as entries 16, 17 and 24,
  differing only in a trailing `.` or `,`.

TraitMech reached the same verdict over a far larger sample (their #249: 353
sidecars, 194 broken markdown-link tails, 2,770 stray trailing commas, 332 of 353
duplicating a reference).

The contract, now written down in `docs/RESEARCH_ARTIFACT_CONTRACT.md`:

| artifact | status |
|---|---|
| References section inside the report `.md` | **authoritative** |
| client `--separate-citations` sidecar | **disabled** |
| `<stem>-citations.md` (Edison path, our own parser) | **derived**, for skimming |

This drops a broken duplicate, not a source. The one pre-existing sidecar stays
on disk as the evidence for the decision, and the doc says explicitly that it
must not be read as a citation list.

## 2. Focus now selects the prompt

`conf/deep_research_provider.yaml` ranks providers per focus, but the justfile
pinned `--template media_growth_research.md` unconditionally. So
`--focus formulation` would rank providers for formulation work and then research
growth evidence anyway — the triage/dispatch disconnect #289 reports.

| focus | template |
|---|---|
| `growth_evidence` (default) | `templates/media_growth_research.md` |
| `formulation` | `templates/media_recipe_validation.md` |

`--template` still overrides, so the prompts that are steps in a named workflow
rather than standing focuses — `media_stock_solution_research.md` (#150 cocktail
repair), `media_axis_classification.md`, `medium_organism_recipe_extraction.md`
(phase 2) — stay reachable without inventing a focus for each.

## 3. Output label carries the focus

`research/media/<category>/<slug>-deep-research-<focus>-<provider>.md`, including
for the default focus, so a caller can predict the path from (slug, focus,
provider) alone without knowing which focus is default — and two focuses cannot
collide on one filename. The single pre-existing report under the old name is
left alone.

## 4. Shared runner interface

```
just research-entity <provider> <target> [focus]   # the fleet contract shape
just research-media  <provider> <target> [focus]   # compatibility alias, identical
just research-focuses                              # prints the table from the code
```

The alias is kept because scripts, batch files and the skills call
`research-media`.

## Verification

```
$ just research-focuses
formulation: templates/media_recipe_validation.md
growth_evidence (default): templates/media_growth_research.md

$ just research-media claude_code ko2_no3 formulation --dry-run
Researching: ko2_no3 (claude_code, focus=formulation) -> research/media/bacterial/ko2_no3-deep-research-formulation-claude_code.md
Template: templates/media_recipe_validation.md
Citations: in-report References section (no separate sidecar)
```

11 new tests: no citation sidecar is ever requested; every focus maps to a
template that exists; the focuses map to *distinct* templates (the actual #289
defect); focus and provider both reach the filename; unknown focus is rejected at
the CLI; explicit `--template` still overrides.

**1,221 tests pass, 2 skipped. Ruff clean. No provider was called and no credits
were spent** — verification is dry-run only.

## Scope

#289 is a fleet issue. This PR does CultureMech. The remaining items —
MediaIngredientMech and CommunityMech still request `--separate-citations`, and
the shared contract tests across repositories — are changes in other
repositories and will be filed there with this as the reference. proteintraitsmech
has no deep-research runner on GitHub at all, so #289's "five generic runners" is
four.

### DIFF
--- a/.claude/skills/deep-research-medium/SKILL.md
+++ b/.claude/skills/deep-research-medium/SKILL.md
@@ -177,8 +177,11 @@ Outputs (per task; `<stem>` is `<slug>-edison-literature`):
 - `<stem>-response.json` — full
   `response.model_dump(mode="json")` plus the verbose-fetch dump.
   Every SDK-exposed field, future-proof against new ones.
-- `<stem>-citations.md` — parsed reference list from the answer
-  (numbered entries with PMID / DOI / URL extracted).
+- `<stem>-citations.md` — **derived, not authoritative.** A local
+  best-effort parse of the answer (numbered entries with PMID /
+  DOI / URL extracted), for skimming only. The authoritative
+  citation record is the References section inside `<stem>.md`;
+  cite from there. See `docs/RESEARCH_ARTIFACT_CONTRACT.md`.
 - `<stem>-agent-state.json` — PaperQA's `agent_state` (tool-call
   trace), `environment_frame`, and verbose `metadata`. Only
   written when the verbose fetch returns any of those.
--- a/docs/RESEARCH_ARTIFACT_CONTRACT.md
+++ b/docs/RESEARCH_ARTIFACT_CONTRACT.md
@@ -18,7 +18,19 @@ just research-entity <provider> <target> [focus] [-- extra args]
 - **focus** — one of the focuses below. Selects the prompt template *and* the
   output label. Defaults to `growth_evidence`.
 
-`just research-media` is a compatibility alias with identical behaviour.
+`focus` is positional here, so **flags must come after it**:
+
+```
+just research-entity claude_code ko2_no3 formulation --dry-run
+```
+
+`just research-media <provider> <target> [flags]` is the compatibility alias.
+Its signature is deliberately *not* the same: it takes **no positional focus**,
+because adding one would bind `--dry-run` in
+`just research-media claude_code ko2_no3 --dry-run` to the focus. Existing
+callers keep working; pass `--focus formulation` through its flags if you need a
+non-default focus, or use `research-entity`.
+
 `just research-focuses` prints the table below from the code.
 
 ### Focuses
--- a/docs/RESEARCH_ARTIFACT_CONTRACT.md
+++ b/docs/RESEARCH_ARTIFACT_CONTRACT.md
@@ -1,0 +1,105 @@
+# Deep-research artifact contract
+
+What each file a research run leaves behind *means*, and which ones may be cited
+as evidence. Introduced by #289, which found that the Mechs disagreed about
+citation artifacts and about how a runner is invoked.
+
+## The entity-runner contract
+
+```
+just research-entity <provider> <target> [focus] [-- extra args]
+```
+
+- **provider** — `claude_code`, `openscientist`, `falcon`, `cborg`, … Aliases
+  (`edison` → `falcon`) resolve in `scripts/deep_research_provider.py`.
+- **target** — a media YAML path, filename slug, `CultureMech:NNNNNN` id, or
+  record `name`. Resolution is the domain-specific part; see
+  `resolve_media_file`.
+- **focus** — one of the focuses below. Selects the prompt template *and* the
+  output label. Defaults to `growth_evidence`.
+
+`just research-media` is a compatibility alias with identical behaviour.
+`just research-focuses` prints the table below from the code.
+
+### Focuses
+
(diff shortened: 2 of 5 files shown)


realmarcin:
Adversarial review pass — found one real defect in my own change, fixed in 487cbfc.

**The compatibility alias was not compatible.** Giving `research-media` a positional `focus` meant the documented form

```
just research-media claude_code ko2_no3 --dry-run
```

bound `--dry-run` to `focus`. The runner then received `--focus --dry-run`, failed argparse with `expected one argument`, and never ran the dry run. Every existing caller that passes a flag directly after the target would have broken — including the form in `prompts/claude_code_deep_research.md`.

`research-media` is back to its original `provider target *args` signature and forwards the default focus; a non-default focus reaches it as `--focus formulation` through args (argparse takes the last value, so the forwarded default loses to an explicit one). `research-entity` keeps the positional focus — that is the contract shape #289 specifies and it has no existing callers — with the ordering constraint stated in the recipe comment.

`--focus={{focus}}` instead of `--focus {{focus}}` makes the remaining footgun legible:

```
before: argument --focus: expected one argument
after:  argument --focus: invalid choice: '--dry-run' (choose from 'formulation', 'growth_evidence')
```

Pinned by `test_the_research_media_alias_takes_no_positional_focus`, which reads both signatures out of `project.justfile`. Verified it fails when the positional focus is restored, and passes when it is not.

All four invocation forms re-verified by dry run. **1,222 tests pass, 2 skipped; no provider called, no credits spent.**

Also checked and found clean:
- No other caller of `build_command` or `citations_file` outside the runner and its tests, so dropping the `citations_file` parameter breaks nothing.
- The pre-existing report `research/media/algae/2asw-deep-research-falcon.md` and its sidecar are untouched by the filename change — nothing globs that pattern outside tests.

realmarcin:
Second adversarial pass (post-review of the review). One real defect in this PR, fixed in 1c9ceba; one pre-existing defect found and filed.

**Fixed here — unquoted recipe expansions.** Record `name:` values contain spaces and a target may be one. `{{target}}` expanded unquoted, and the alias expanded it a *second* time through the recursive `just` call, so `just research-media claude_code "LB Broth"` split into `target=LB`, `focus=Broth`:

```
error: argument --focus: invalid choice: 'Broth'
```

Quoted, the resolver receives it whole and gives the answer it should:

```
ValueError: Target 'LB Broth' matched multiple media records equally specifically:
  data/normalized_yaml/bacterial/TOGO_M3227_LB_broth.yaml,
  data/normalized_yaml/bacterial/lb_broth.yaml
```

`{{args}}` stays unquoted — it is a word list, and quoting it would collapse `--max-cost 1` into one argument. Pinned by `test_the_recipes_quote_the_values_that_can_contain_spaces`.

**Filed, not fixed — #297.** `--max-cost` never reaches the client: `argparse.REMAINDER` on a positional claims a leading-`--` token as an unknown option first, so every `deep-research-client` flag is unreachable through the runner. Confirmed pre-existing by running main's arg shape and this branch's side by side — byte-identical failure:

```
main:   ... --template ... --max-cost 1  -> error: unrecognized arguments: --max-cost
branch: ... --focus=... --max-cost 1     -> error: unrecognized arguments: --max-cost
```

The existing tests miss it because they call `build_command` directly, where passthrough works; the break is purely CLI routing. Likely affects the sibling runners too, which matters for #289's "one contract" goal.

Also re-checked and clean: no caller outside this script imports `build_command` or `DEFAULT_TEMPLATE` from `research_media` (`research_media_edison.py`, `render_media_prompt.py`, and `research_organism_recipe_edison.py` each define their own), so dropping the `citations_file` parameter and rebasing the template default break nothing.

CI green: pytest, validate, consistency, plausibility.

Issue #283, opened 2026-08-17
ucm.yaml lists the same trace elements twice, 1000x apart — stock strength and final concentration side by side


realmarcin:
Found probing #282 for rename-induced collisions. It is **not** caused by that PR — the
same rows exist on main under the corrupted spellings — but repairing the names made it
legible.

`data/normalized_yaml/algae/ucm.yaml` (and its merged twin) carries each of two trace
elements **twice**, at values exactly 1000× apart:

| ingredient | value A | value B | ratio |
|---|---|---|---|
| `As2O3` (was `As O`) | 0.000093 g/L | 0.093 g/L | 1000× |
| `Na2WO4.2H2O` (was `Na WO .2H O`) | 0.000066 g/L | 0.066 g/L | 1000× |

Both rows in each pair carry the same grounding and the same source
(`CCAP Medium MR_UCM.pdf`).

A factor of exactly 1000 between two copies of one ingredient is the **flattened-cocktail
signature** from #150: the stock-strength value and the final per-litre value both present
as sibling ingredient rows, rather than the stock nested under a `solutions:` entry with an
addition volume. Here it has not been flattened *into* the medium — both readings are
sitting next to each other, which is arguably worse, because any consumer summing
ingredients counts the trace element at ~1000× its real contribution.

Worth checking against the CCAP source PDF which figure is the medium's and which is the
stock's, then either nesting the stock (the #150 repair) or dropping the stock-strength
row.

Also worth a corpus-wide sweep for the same shape — two rows, one name, values differing
by a clean factor of 1000 — since this one only surfaced by accident. The existing
`audit_concentration_plausibility` looks at absolute magnitude, not at ratios between
sibling rows, so it would not catch this pattern.

Related: #263 removed rows duplicated *verbatim*; this is the sibling case where the
duplicate carries a different value and so is a real (wrong) datum rather than redundancy.

realmarcin:
**This is understated — it is 8 duplicated ingredients, not 2, and there are
three distinct patterns.** Re-measured on `main` at `e75a797bbf`,
`data/normalized_yaml/algae/ucm.yaml` (`CultureMech:000140`, 47 ingredients):

| ingredient | values (g/L) | ratio |
|---|---|---|
| `Na2SeO3 x 5 H2O` | 0.00263 / 2.63 | **1000x** |
| `As2O3` | 0.000093 / 0.093 | **1000x** |
[...]

Issue #273, opened 2026-08-15
Preparation instructions are parsed as ingredients with concentrations


realmarcin:
Noticed while canarying #263's dedup.

`data/normalized_yaml/algae/sbbm.yaml` carried this as an ingredient row:

```yaml
- preferred_term: Make up to 1 litre with distilled water. For agar add
  concentration:
    value: '15'
    unit: G_PER_L
  source: CCAP Medium MR_SBBM.pdf
```

That is a preparation instruction from the source PDF, parsed as an ingredient and given
the concentration of whatever followed it (the 15 g/l is presumably the agar). It appeared
twice, which is how the dedup pass surfaced it.

Two distinct harms:

1. **A phantom ingredient.** Anything counting ingredients, exporting to KG-Microbe, or
   computing composition sees a component that does not exist.
2. **A misattributed concentration.** If the 15 g/l belongs to agar, then agar is either
   missing its concentration or carrying a different one — worth checking which.

This is likely not one record. Candidate detection: `preferred_term` containing a verb
phrase, sentence punctuation, or exceeding some length — real ingredient names are short
noun phrases. A quick scan for `preferred_term` values containing `. ` or more than ~6
words would size it.

Related: #265 (ingredient synonyms effectively unpopulated) and #259 (`term.label` carrying
ingredient strings) are both symptoms of the same underlying thing — the ingredient field
is absorbing text it was not meant to hold.

realmarcin:
**Do not implement this as a deletion — I checked, and it would lose real ingredient data.**

Scoping it properly. A conservative pattern (imperative volume/pH/dissolution phrases only,
excluding chemical formula notation like `MgSO4 . 7H2O` and genuine ingredients like
`Sterile Defibrinated Rabbit Blood`) finds **86 rows across 22 distinct strings**.

**All 86 carry a concentration.** That is the important part, and it is not incidental — it
confirms the parser attached the *following* ingredient's quantity to the instruction text.

`data/normalized_yaml/algae/sbbm.yaml` in full:

| preferred_term | value |
|---|---|
| `Make up to 1 litre with distilled water. For agar add` | 15 G_PER_L |
| `Add to 1000 ml of distilled water` | 0.75 G_PER_L |
| FeCl3 x 6 H2O | 0.097 |
| … | |
| `Soil is prepared as above.` | 105 G_PER_L |

**There is no separate agar ingredient in that record.** The 15 g/l agar exists *only* on
[...]

Issue #271, opened 2026-08-15
MIM and CultureMech both split L-cysteine HCl monohydrate by punctuation: 'x H2O' vs '-HCl x H2O'


realmarcin:
Found reviewing #270, and it is a two-sided finding — both repos have the same defect,
so a fix on either side alone leaves the split in place.

`L-cysteine hydrochloride monohydrate` is written several ways across the corpus. MIM's
`UNIFIED_INGREDIENT_MAPPING.tsv` grounds those spellings to **two different terms**, split
by nothing but punctuation:

| ingredient name | MIM `mim_id` | is |
|---|---|---|
| `L-Cysteine HCl x H2O` (space) | CHEBI:91248 | L-cysteine hydrochloride hydrate ✅ |
| `L-Cysteine・HCl・H2O` | CHEBI:91248 | ✅ |
| `L-cysteine HCl.H2O` | CHEBI:91248 | ✅ |
| `L-Cysteine hydrochloride monohydrate` | CHEBI:91248 | ✅ |
| **`L-Cysteine-HCl x H2O`** (hyphen) | **CHEBI:17561** | plain L-cysteine ❌ |
| **`Cysteine-HCl x H2O`** | **CHEBI:17561** | ❌ |
| **`Cysteine-HCl·H2O`** | **CHEBI:17561** | ❌ |
| **`L-cysteine-HCL x H2O`** | **CHEBI:17561** | ❌ |

Same substance, same source, and the only difference is whether `HCl` is preceded by a
space or a hyphen. CHEBI:17561 is neither the salt nor the hydrate.

CultureMech had exactly the same split — #270 fixes our side, moving the hyphenated forms
to CHEBI:91248 to match what both repos already do for the spaced forms (1,901 rows here,
and MIM's own mapping above).

**Consequence for the divergence count.** After #270, 63 ingredient rows have
`culturemech_term_id` = CHEBI:91248 against MIM's `mim_id` = CHEBI:17561. Per
MediaIngredientMech#138's precedence chain, ours wins downstream — so kg-microbe gets the
right answer. But the divergence is real and will show up in any future join, and MIM's
copy stays wrong for anyone reading MIM directly.

Worth reporting upstream so MIM's hyphenated variants move to CHEBI:91248 too. That is a
MIM-side edit; I have not filed it there.

Related: #258 (one ingredient name, several CHEBI ids) is the same disease in a different
compound family — the general fix is a normalization step that collapses punctuation
[...]

PR #295, opened 2026-08-19
Make the KGX export run, emit TSV, and declare the nodes it mints (#294)

### LINKED ISSUE #294: just kgx-export cannot run, emits no TSV, and mints six node types it never declares
`just kgx-export` cannot run, and even when it does it cannot produce a KGX TSV
pair. Three separate defects in the same path. Found while triaging which open
issues are critical to the KGX TSV export — none of them cover this, because the
export has not been exercised.

## 1. The recipe cannot spawn koza, and passes the wrong argument type

```
$ just kgx-export
error: Failed to spawn: `koza`
  Caused by: No such file or directory (os error 2)
```

`koza` is an optional extra in `pyproject.toml` (`[koza]`, line 71), but
`project.justfile:1560` runs `uv run koza ...` with no `--extra koza`. It is not
in the default or `dev` environments:

```
$ uv run python -c "import koza"              -> ModuleNotFoundError
$ uv run --extra dev python -c "import koza"  -> ModuleNotFoundError
$ uv run --extra koza koza --help             -> works
```

Adding the extra is necessary but not sufficient. The recipe calls:

```
uv run koza transform src/culturemech/export/kgx_export.py -i ... -o ... -f jsonl
```

In koza 2.1.1 the positional argument to `transform` is a **configuration YAML**,
not a Python file:

```python
def transform(
    configuration_yaml: Annotated[str, typer.Argument(help="Configuration YAML file")],
    ...
```

There is no koza config YAML anywhere in the repo. So the recipe has never run
against this koza version, and `tests/test_kgx_export.py` passes only because it
exercises the pure `transform()` function, never the koza path.

## 2. It emits jsonl, so there is no TSV product

`-f jsonl` is hardcoded. `-f tsv` is the format that writes the KGX
`<name>_nodes.tsv` / `<name>_edges.tsv` pair.

Note `-f kgx` is a trap — the value is in `OutputFormat` but `runner.py` has no
branch for it, so it falls through to `raise ValueError("No writer defined")`.
`tsv` is the correct choice.

Also: `TSVWriter` only creates a file when the corresponding property list is
configured:

```python
if node_properties:  # Make node file
if edge_properties:  # Make edge file
```

Both default to `None` in `WriterConfig`, so a config that omits them produces
**no output files at all**.

## 3. The transform emits edges only — every minted node dangles

`_make_association` returns a plain dict with `subject`/`predicate`/`object` and
no `category`, and the koza wrapper only ever writes `Association` objects. No
node record is produced anywhere.

That is fine for `CHEBI:`, `NCBITaxon:`, `MICRO:`, `FOODON:` and `TOGO:` objects —
those nodes come from KG-Microbe's ontology ingests. It is not fine for the six
id shapes CultureMech mints itself:

| minted id | source |
|---|---|
| `culturemech:<name>` | the medium (line 102) |
| `culturemech:medium_type_<TYPE>` | line 380 |
| `culturemech:application_<slug>` | line 461 |
| `culturemech:state_<state>` | line 480 |
| `culturemech:<variant>` | line 541 |
| `culturemech:solution_<slug>` | line 594 |

Nothing else in the fleet defines them, so they are dangling references in any
`nodes.tsv`. Confirmed on one record:

```
$ transform(lb_broth.yaml)
7 edges
id prefixes: {culturemech: 4, TOGO: 1, CHEBI: 1, MICRO: 1, FOODON: 1}
```

Four of the five distinct ids in a single record are CultureMech-minted and
node-less.

## Proposed fix

1. Add a koza configuration YAML with the reader, the transform module, and
   explicit `node_properties` / `edge_properties`.
2. Run it as `uv run --extra koza koza transform <config> -f tsv`.
3. Emit node records for the six minted id shapes only, leaving ontology terms to
   the ontology ingests. Align categories with the consumer — kg-microbe already
   fixes these in `kg_microbe/transform_utils/constants.py`:

   ```
   MEDIUM_DEFINED_CATEGORY  = "biolink:GrowthMedium|biolink:ChemicalMixture"
   MEDIUM_COMPLEX_CATEGORY  = "biolink:GrowthMedium|biolink:ComplexMolecularMixture"
   MEDIUM_TYPE_DEFINED_CATEGORY = "biolink:ChemicalMixture"
   MEDIUM_TYPE_COMPLEX_CATEGORY = "biolink:ComplexMolecularMixture"
   SOLUTION_CATEGORY = "biolink:ChemicalMixture"
   ```

   `biolink:GrowthMedium` is not in the installed `biolink_model` pydantic model
   and its classes pin `category` to a per-class literal, so node records need to
   be plain dataclasses rather than biolink pydantic objects. koza supports this —
   `KGXConverter.convert_node` falls back to `asdict()` for non-`BaseModel`
   entities, and `split_entities` classifies anything with `id` and `name` (and no
   subject/predicate/object) as a node.
4. Deduplicate shared nodes within a run — `culturemech:medium_type_COMPLEX`
   would otherwise be emitted once per record across 15,878 records.
5. Add a test that exercises the koza path end to end on a small fixture set, so
   this class of breakage cannot pass CI again.

## Acceptance criteria

- `just kgx-export` runs to completion from a clean checkout.
- It writes a KGX node/edge TSV pair.
- Every `culturemech:`-prefixed id appearing in the edges file has a row in the
  nodes file.
- Node categories match what kg-microbe expects.
- The koza path — not just the pure `transform()` function — is covered by a test.

realmarcin:
Closes #294.

`just kgx-export` could not run at all. Running it exposed a fourth defect worse
than the three the issue names, and the full-corpus run exposed a fifth the
canary could not reach.

## What was broken

**1. It never ran.** `koza` is an optional extra; the recipe called `uv run koza`
with no `--extra koza` and died at `Failed to spawn`. It also passed
`kgx_export.py` where koza 2.x expects a **configuration YAML**, and no koza
config existed in the repo. `tests/test_kgx_export.py` passes because it
exercises only the pure `transform()` function.

**2. No TSV was possible.** `-f jsonl` was hardcoded. `-f tsv` writes the KGX
pair. `-f kgx` is a trap — the enum value exists but `runner.py` has no branch
and falls through to `No writer defined`. `TSVWriter` also only creates a file
when the matching property list is configured; both default to `None`.

**3. Every minted node dangled.** Edges only, so the six `culturemech:` id shapes
we mint were declared nowhere.

**4. 99% of edges were being dropped — found by the canary.** Biolink's
`Association.qualifiers` is `list[str] | None`, so `Association(**edge_dict)`
raised for every edge carrying a qualifier dict, and the wrapper swallowed it
with a `print`:

```
249 records -> 10 edges     (before)
249 records -> 1,937 edges  (after)
```

**5. Node and edge ids disagreed — found only on the full corpus.** Koza's
`trim()` strips the sequence `\"` from every edge column, but
`TSVWriter.write_row` restores a node's `id` from the raw record and bypasses it.
Two records carry literal backslash-quotes in a solution name:

```
node: culturemech:solution_Mineral_salt_solution*_\"Hutner_Cohen-Bazire\"
edge: culturemech:solution_Mineral_salt_solution*_Hutner_Cohen-Bazire
```

2 dangling references, 2 orphan nodes. `_sanitize_id` now drops quotes,
backslashes and asterisks at the source so both sides agree regardless of koza's
asymmetry, and the medium id goes through it too instead of only replacing
spaces.

## What changed

- `scripts/export_kgx.py` — driver. Koza's CLI wants a config YAML, does no glob
  expansion, and 15,878 paths overflow ARG_MAX (`ls data/normalized_yaml/*/*.yaml`
  already fails), so the file list goes through koza's Python API. Clears the
  run-scoped node-dedup set and **fails loudly** on a missing or header-only
  output file — a silent empty export is the failure mode here.
- `src/culturemech/export/kgx.yaml` — the koza config, with explicit
  `node_properties` / `edge_properties`.
- `nodes()` + `Node`/`Edge` dataclasses. Categories come from the consumer, not
  invented: kg-microbe's `MEDIUM_DEFINED_CATEGORY`, `MEDIUM_COMPLEX_CATEGORY`,
  `SOLUTION_CATEGORY`. `BUFFER` / `NEGATIVE_CONTROL` fall back rather than
  guessing at a composition they do not assert. The module no longer imports
  biolink at all.
- `just kgx-export` rewritten, plus `just kgx-export-sample <category>` so the
  next person can canary before committing to a 15,877-record run.

## Verification — canary first, then the full corpus

```
$ just kgx-export-sample algae      # 249 records
  259 nodes, 1,937 edges

$ just kgx-export                   # 15,877 records, 1m49
  output/kgx/culturemech_nodes.tsv:   9,314 rows
  output/kgx/culturemech_edges.tsv: 198,010 rows

  culturemech ids referenced 9,314 — DANGLING 0, ORPHANED 0
  qualifiers kept: concentration 153,066, attribute_type 34,246, strain 53
  external prefixes (correctly not declared here): komodo.medium 3,637,
    mediadive.medium 3,325, TOGO 2,917, CHEBI 572, MEDIADB 469, NCBITaxon 73
```

13 new tests drive the **real koza runner** over fixture records and assert on
the files on disk — the coverage gap that let all five defects ship. Includes the
acceptance criterion (no dangling `culturemech:` id), the qualifier regression,
and the id-drift case.

**1,226 tests pass, 2 skipped.** `output/` is gitignored, so no artifacts are
committed.

## Note on scope

This makes the export *run and be internally consistent*. It does not fix the
record-level data defects that feed it — #256, #273, #279, #283, #150, #271 are
the ones that put wrong or phantom nodes into the graph, and they are now
verifiable against real output for the first time.

### DIFF
--- a/project.justfile
+++ b/project.justfile
@@ -1523,16 +1523,34 @@ generate-all-indexes:
 # EXPORT
 # ================================================================
 
+# Export the recipe corpus as a KGX node/edge TSV pair (#294).
+#
+# Driven by scripts/export_kgx.py rather than a bare `koza transform` because
+# koza's CLI takes a config YAML (not a .py), does no glob expansion, and the
+# 15,878-file corpus overflows ARG_MAX. The script also clears the run-scoped
+# node-dedup set and fails loudly on an empty or missing output file.
+#
+# `--extra koza` is required: koza is an optional dependency and is in neither
+# the default nor the `dev` environment.
 [group('Export')]
-kgx-export:
+kgx-export *args="":
+    uv run --extra koza python scripts/export_kgx.py \
+      --records-dir {{normalized_yaml_dir}} \
+      --output-dir {{output_dir}}/kgx \
+      {{args}}
+
+# Canary the export on one category before committing to the full corpus.
+# Example: just kgx-export-sample algae
+[group('Export')]
+kgx-export-sample category="algae":
     #!/usr/bin/env bash
-    mkdir -p {{output_dir}}/kgx
-    echo "Exporting recipes to KGX format..."
-    uv run koza transform src/culturemech/export/kgx_export.py -i '{{normalized_yaml_dir}}/**/*.yaml' -o {{output_dir}}/kgx -f jsonl
-    echo ""
-    echo "✓ KGX edges exported to {{output_dir}}/kgx/"
-    echo "Stats:"
-    wc -l {{output_dir}}/kgx/*.jsonl || echo "No output files generated"
+    set -euo pipefail
+    tmp="$(mktemp -d)"
+    trap 'rm -rf "$tmp"' EXIT
+    cp -r "{{normalized_yaml_dir}}/{{category}}" "$tmp/"
+    uv run --extra koza python scripts/export_kgx.py \
+      --records-dir "$tmp" \
+      --output-dir {{output_dir}}/kgx-sample
 
 # ================================================================
 # BROWSER
--- a/scripts/export_kgx.py
+++ b/scripts/export_kgx.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python3
+"""Run the CultureMech KGX export through koza and write a node/edge TSV pair.
+
+Why a driver script rather than a bare `koza transform` call (#294):
+
+1. koza's CLI takes a **configuration YAML**, not a Python file. The old recipe
+   passed `kgx_export.py`, so it had never run against koza 2.x.
+2. koza does **no glob expansion**, and the corpus is 15,878 separate YAML
+   records. The expanded list overflows ARG_MAX — `ls data/normalized_yaml/*/*.yaml`
+   already fails — so the file list cannot go through `-i` on the command line.
+   Passing it through koza's Python API sidesteps the shell entirely.
+3. The node-dedup set is run-scoped and has to be cleared before each run.
+
+Reports node and edge counts, and fails loudly if either file is missing or
+holds only its header — a silent empty export is the failure mode #294 was
+about.
+"""
+from __future__ import annotations
+
+import argparse
+import sys
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+DEFAULT_RECORDS = REPO_ROOT / "data" / "normalized_yaml"
+DEFAULT_OUTPUT = REPO_ROOT / "output" / "kgx"
+CONFIG = REPO_ROOT / "src" / "culturemech" / "export" / "kgx.yaml"
+
+
+def find_records(records_dir: Path) -> list[str]:
+    """Every record YAML under `records_dir`, one directory deep.
+
+    Matches the corpus layout `data/normalized_yaml/<category>/<slug>.yaml`. The
+    sibling `*_index.json` files live at the top level and are not YAML, so the
+    `*/*.yaml` shape excludes them without needing a filter.
+    """
+    return [str(p) for p in sorted(records_dir.glob("*/*.yaml"))]
+
+
+def run(records_dir: Path, output_dir: Path, limit: int = 0) -> tuple[int, int]:
+    """Run the transform. Returns (node_count, edge_count) excluding headers."""
+    from koza.model.formats import OutputFormat
+    from koza.runner import KozaRunner
+
+    sys.path.insert(0, str(REPO_ROOT / "src"))
+    from culturemech.export.kgx_export import reset_node_dedup
+
+    files = find_records(records_dir)
+    if not files:
+        raise SystemExit(f"No record YAMLs found under {records_dir}")
+
+    # Koza imports the transform module itself, so the dedup set has to be
+    # cleared through the same module object koza will use, not a fresh import.
+    reset_node_dedup()
+
+    print(f"Reading {len(files)} records from {records_dir}")
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    _config, runner = KozaRunner.from_config_file(
+        str(CONFIG),
+        input_files=files,
+        output_dir=str(output_dir),
+        output_format=OutputFormat.tsv,
+        row_limit=limit,
+    )
+    runner.run()
+
+    return _verify(output_dir)
+
+
+def _verify(output_dir: Path) -> tuple[int, int]:
+    """Confirm both files exist and carry rows. Raises SystemExit otherwise."""
+    counts = []
+    for kind in ("nodes", "edges"):
+        path = output_dir / f"culturemech_{kind}.tsv"
+        if not path.exists():
+            raise SystemExit(
+                f"{path} was not written. TSVWriter only creates a file when the "
+                f"matching property list is set in {CONFIG.name}."
+            )
+        rows = sum(1 for _ in path.open()) - 1  # discount the header
+        if rows <= 0:
+            raise SystemExit(f"{path} has a header but no rows.")
+        counts.append(rows)
+        print(f"  {_display(path)}: {rows:,} rows")
+    return counts[0], counts[1]
+
+
+def _display(path: Path) -> str:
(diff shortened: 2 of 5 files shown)


realmarcin:
Adversarial review pass — one real defect found in my own change, fixed in 33a20dc.

**19 minted ids carried a second colon.** A solution named `Solution A:` produced `culturemech:solution_Solution_A:`. The colon is the CURIE delimiter, so any consumer that splits on it reads those ids wrongly — and this export exists to be consumed by kg-microbe. Mapped to an underscore rather than dropped, so `...ferrous sulfate:Add 13.9 g/l` does not weld two words together.

Full corpus re-run after the fix:

```
9,306 nodes (down from 9,314 — fixed spellings merge), 198,010 edges
DANGLING 0   ORPHANED 0   ids with >1 colon: 0
```

Also checked and found clean:

- **No whitespace or control characters in any id.** This mattered specifically because `TSVWriter.write_row` writes a node `id` raw, bypassing koza’s `trim()` — a tab in an id would have silently corrupted the TSV. Zero occurrences.
- **No id maps to two different names**, so the run-scoped dedup (which keeps the first row for an id) is not masking a conflict.
- **`output/` is gitignored**, so the 198k-row artifacts are not in the diff.
- **`_verify`’s line count is a valid row count** given the above — no embedded newlines.

Two things deliberately left out and filed instead:

- **#296** — local ids still carry `[`, `]`, `%`, `,` and, more concerning, three different middle-dot characters (`·` 304, `・` 80, `•` 1) plus straight vs curly apostrophes. These parse, so nothing is broken, but two names differing only in which dot they use become two distinct nodes. Wants a collision measurement before picking a normalization, which is more than this PR should decide.
- **#273** — the longest minted id in the corpus is an entire preparation protocol sitting in a solution `preferred_term`. That is #273’s defect on the `solutions` side rather than `ingredients`; commented there with the evidence and a suggestion to widen the detection sweep. Fixing it at the record level is what removes the ugliest ids, not a sanitizer change.

**1,226 tests pass, 2 skipped.** Full corpus re-verified after every change, canary (`algae`, 249 records) before each full run.

realmarcin:
Second adversarial pass. One real defect in my own change, fixed in 6b4f53b.

**`reset_node_dedup()` was dead code carrying a false guarantee.** The driver called it with a comment claiming it cleared "the same module object koza will use". It does not. Koza loads the transform with `importlib.util.spec_from_file_location` and — in its own source comment — "without touching sys.modules", so it builds a **fresh module object with a fresh, empty `_EMITTED_NODE_IDS` on every run**. The driver was clearing the copy imported as `culturemech.export.kgx_export`, which koza never sees.

Proved rather than reasoned:

```
run1: 3 nodes 2 edges
run2: 3 nodes 2 edges
package-module dedup set after both runs: set()   <- never touched
```

The behaviour was correct by accident. That is worse than no code, because the next person relies on the stated guarantee. Dropped the call, documented the real mechanism, and pinned the property that actually matters — two runs in one process produce byte-identical output (`test_a_second_run_in_the_same_process_repeats_the_output`). If koza ever starts caching transform modules, that test fails and `reset_node_dedup()` becomes the ready-made fix; the function is kept for that reason and for callers driving `nodes()` directly.

Also re-checked on the full 15,877-record corpus and clean:

- **0 dangling, 0 orphaned** `culturemech:` ids; **0 ids with more than one colon**.
- **No whitespace or control characters in any id** — which mattered specifically because `TSVWriter.write_row` writes a node `id` raw, bypassing koza’s `trim()`; a tab would have silently corrupted the TSV.
- **No id maps to two different names**, so the dedup (which keeps the first row per id) is not masking a conflict.
- `output/` is gitignored, so no 198k-row artifacts in the diff.

Two things deliberately filed instead of fixed: **#296** (confusable characters — three distinct middle dots — which needs a collision measurement before choosing a normalization) and a comment on **#273** (the longest minted id is an entire preparation protocol in a solution `preferred_term`, i.e. #273’s defect on the solutions side).

CI green: pytest, validate, consistency, plausibility.

Issue #299, opened 2026-08-19
NBRC composition tables were never parsed: chemical names sitting in concentration.value, whole recipes in solution names


realmarcin:
Found while sizing #273 against the now-runnable KGX export (#294). This is a
**different and more severe defect** than #273, which is why it is filed
separately: #273 is prose leaking into a name field; this is a whole composition
table that was never parsed, with the chemical name landing in the *concentration*.

## The signature

`data/normalized_yaml/bacterial/NBRC_1003.yaml`:

```yaml
ingredients:
- preferred_term: ''                    # <- name is empty
  concentration:
    value: MgSO4·7H2O                   # <- the name is HERE
    unit: G_PER_L
solutions:
- preferred_term: MgSO4·7H2O0.5g(NH4)2SO40.4gK2HPO40.2gKCl0.1gDistilled water1LAdjust
    pH to 2.0 with H2SO4.Heterotrophic growth on yeast extract:Add 10 mg/l of FeSO4·7H2O
    to the medium. After autoclaving, add yeast extract from sterile stock solution
    to final concentration of 0.25 g/l. ...
  composition: []                       # <- the actual composition is empty
  name: Unknown solution
```

The source table `MgSO4·7H2O 0.5g / (NH4)2SO4 0.4g / ...` was concatenated into
one string and dropped into `solutions[].preferred_term`, `composition` was left
empty, and a single ingredient row was emitted whose name/value are swapped. The
record's composition is unrecoverable from the normalized layer.

## Scale, measured over all 15,877 records

| signature | rows | records |
|---|---:|---:|
| `ingredients[].preferred_term` is empty | 94 | 94 |
| `ingredients[].concentration.value` is non-numeric | 1,616 | 593 |
| both — the name landed **in** the concentration | 59 | 59 |
| `solutions[]` with `composition: []` and a concatenated-table name | 37 | 37 |

169,512 ingredient rows scanned. The 1,616 non-numeric concentrations are the
widest band and likely mix several causes; the 59 name-into-concentration swaps
are unambiguous.

Concentrated in NBRC imports (`NBRC_1003`, `1021`, `1038`, `1051`, `1063`, …),
which points at one importer rather than scattered curation error.

## It does reach the KGX export

Unlike #273's ungrounded ingredient rows — which produce **0 nodes and 0 edges**,
because `medium_to_ingredient_edge` returns `None` without a `term.id` — solution
ids are minted from `preferred_term` and need no grounding. Measured on the full
export:

[...]

PR #300, opened 2026-08-19
Detect and gate composition tables that were never parsed (#299, #273)

realmarcin:
Addresses the detection half of #299 and #273, and corrects a claim I made about #273.

## What sizing #273 actually found

Two different defects were sitting in the same scan.

**#273's phantom ingredients never reach the KG.** I had listed #273 as putting phantom nodes and edges into the export. That is wrong. `medium_to_ingredient_edge` returns `None` without a `term.id`, and all 44 high-confidence prose rows are ungrounded:

```
$ grep -c Make_up_to_1_litre culturemech_nodes.tsv culturemech_edges.tsv
nodes: 0    edges: 0
```

Its harm is to ingredient counts and composition analysis — real, but not the graph.

**#299 does reach the KG.** A family of NBRC imports put the whole composition table into `solutions[].preferred_term` with `composition: []`, and the reagent name into `concentration.value` with an empty `preferred_term`:

```yaml
ingredients:
- preferred_term: ''                     # name is empty
  concentration: {value: MgSO4·7H2O, unit: G_PER_L}   # the name is HERE
solutions:
- preferred_term: MgSO4·7H2O0.5g(NH4)2SO40.4gK2HPO40.2gKCl0.1gDistilled water1L...
  composition: []                        # the composition is gone
```

Solution ids are minted from `preferred_term` and need no grounding, so **all 31 become nodes, one `has_part` edge each** — verified against the export output, not inferred.

## The audit

Four detectors over 15,877 records / 169,512 ingredient rows:

| detector | count | reaches the KG |
|---|---:|---|
| `NAME_IN_CONCENTRATION` | 59 | no |
| `EMPTY_INGREDIENT_NAME` | 35 | no |
| `UNPARSED_SOLUTION_TABLE` | 31 | **yes** |
| `PROSE_AS_INGREDIENT` | 44 | no |

## Precision, because a ratchet punishes false positives

A false positive here raises the baseline and permanently weakens the gate, so both fuzzy detectors were tuned against the corpus rather than guessed:

- **The obvious table test does not work.** "A letter followed by a digit" flags 45, but 14 are legitimate catalogue cross-references — `MINERAL MEDIUM FOR HYDROGENOPHILUS ISLANDICUM (see Medium [M803])` trips on `M803`. Calling those corrupt would be a false accusation. Keying on the actual signature instead — an amount and unit welded onto the next reagent name, the `0.5g(` in `MgSO4·7H2O0.5g(NH4)2SO4` — keeps all 31 tables and drops all 14 references. Stable, too: every real table has ≥4 such runs, so thresholds of 1–4 select the same set.
- **Prose needs a conjunction.** Instruction verb AND long-or-sentence. The loose filters flag 1,192 ingredient values; the conjunction flags 44, and spot-checking says all 44 are real.

## The gate

```
just audit-unparsed-composition          # --max-allowed 169 --max-exported 31
```

Both baselines verified to bite when lowered by one:

```
--max-allowed 168  -> exit 1  FAIL: 169 findings > baseline 168
--max-exported 30  -> exit 1  FAIL: 31 unparsed solution tables > baseline 30
```

`--max-exported` is the sharper of the two: only `UNPARSED_SOLUTION_TABLE` can add garbage nodes to the graph.

## Scope

**Read-only — nothing is repaired here.** Recovery means going back to the pre-normalization payload: the concatenated string has lost the delimiters between reagent and amount (`MgSO4·7H2O0.5g` has no separator), so re-splitting it in this repo would be guesswork on hydrate dots and multi-digit numbers. #299 tracks that, and its first step is checking whether `data/raw/` still holds the NBRC table.

## Notes

- 23 tests, one per detector plus the near-misses that must not trip it.
- `derived_artifacts.tsv` refreshed. The manifest test **correctly failed first** on the new report — the guard working as intended — and picked it up with the right writer attribution.
- 1,268 tests pass, 2 skipped. Ruff clean.


realmarcin:
Adversarial review pass. One real defect found in my own change, fixed in 0f785bf.

**A missing concentration was being read as a swapped name.** An ingredient with an empty `preferred_term` and no concentration block at all reported `NAME_IN_CONCENTRATION` — pointing a curator at a field that does not exist. Same for `concentration: {}`, `value: None`, and `value: ''`:

```
empty name, NO concentration key    -> NAME_IN_CONCENTRATION   (wrong)
empty name, concentration: {}       -> NAME_IN_CONCENTRATION   (wrong)
empty name, value: None             -> NAME_IN_CONCENTRATION   (wrong)
```

Absence is not evidence of a swap — those are `EMPTY_INGREDIENT_NAME`: the name is lost and there is nothing to restore it from.

**Nothing in the corpus is shaped that way today**, so the counts do not move and the committed report is byte-identical. That is precisely why it needed pinning rather than leaving to chance: a wrong detector that currently fires zero times stays invisible until an import trips it, and by then the baseline has been raised around it. In a ratchet, a latent false positive is a slow leak in the gate.

Also inverted `_is_numeric_value` to `_holds_a_name`, so the function answers the question the caller actually asks and absence has one obvious result instead of falling out of a negation.

Also checked and clean:

- **The report is deterministic** — sorted glob, byte-identical across runs, so the committed artifact will not churn on every unrelated PR.
- **Both baselines still bite**, re-verified after the change: `--max-allowed 168` → exit 1, `--max-exported 30` → exit 1.
- **The derived-artifacts manifest guard worked as designed.** It failed on first run because the new report was unregistered, and `just audit-derived-artifacts` picked it up with the correct writer attribution (`audit_unparsed_composition.py`, not a reader). Worth noting given how often derived artifacts drift here.
- **`--max-exported` counts exactly what reaches the graph**: all 31 flagged solution names mint a node id that is present in the export, with 31 edges. Measured against the export output rather than assumed.

27 tests, 1,268 pass overall, ruff clean.

realmarcin:
Third adversarial pass. **Two real defects found in my own change**, both fixed here; two findings filed rather than folded in.

## Fixed — a whole record kind was unscanned

The audit read `ingredients[]` and its scope note claimed "solution RECORDS are not excluded". Both true; together misleading. The **4,784 standalone stock-solution records keep their reagents in a top-level `composition:`**, not `ingredients:` — **35,009 rows the scan never reached** — and reagents nested under `solutions[].composition[]` were missed too.

An audit that looks like full coverage and isn't is worse than one with a stated gap, and in a ratchet the miss is invisible: totals don't move, so nothing signals that a record kind is unguarded. A stock-solution import could land the exact defect this gate exists to catch and pass it.

All three locations now feed the same detectors, with a `location` column so a curator knows which field to open. Counts unchanged — 169 / 31 — which is the useful result: those 35,009 rows are genuinely clean today.

## Fixed — the numeric test rejected scientific notation

`_NUMERIC_VALUE` was `[\d.]+` with an optional range. It does not match `5e-05`, and **the corpus holds 1,781 scientific-notation concentrations — 254 rows of `5e-05` alone**. Any row carrying one of those *and* an empty name would have been reported as a swapped reagent name. Same for a bare `-`, which 72 rows use as a "no value" placeholder.

Same failure mode as the missing-concentration bug from the last pass, and far likelier to fire. Replaced the hand-rolled regex with `float()` — the honest definition, and one that cannot miss a notation the next import happens to use. Ranges now split on a hyphen *between two digits*, so a negative exponent is no longer read as a range separator.

The function is now explicitly biased toward "not a name": a false negative loses one row from a backlog someone is already working; a false positive accuses a valid concentration and gets baked into the baseline. Tests pin **both** directions so the bias cannot drift into missing real swaps.

## Filed, not fixed

- **#301** — 1,781 rows have a *name* and a non-numeric concentration, and nothing looks at them. The band splits at least three ways (scientific notation, `-` placeholders, float noise like `1.0000000000000002e-06`) and needs classifying before it can be gated. Flagging it wholesale would invert this PR's deliberate bias.
- **#302** — the ratchet gates on a **total**, so repairing 10 `PROSE_AS_INGREDIENT` while importing 10 `NAME_IN_CONCENTRATION` nets to 169 and passes. Applies to `audit-concentration-plausibility` too, so it is a fleet-of-audits question. `--max-exported` already narrows it for the one category that reaches the KG.

## Re-verified after every change

```
169 findings, 31 reaching the export      (unchanged through all three fixes)
--max-allowed 168  -> exit 1
--max-exported 30  -> exit 1
report byte-identical across runs
```

47 tests. Ruff clean.

PR #303, opened 2026-08-19
Consume MIM's published SSSOM as a divergence check (#256)

realmarcin:
Item 2 of #256's proposed work: **consume MIM's published SSSOM as a divergence check**, aligned to the latest release.

## Why this has to live here

kg-microbe resolves an ingredient with `best_primary([chebi_id, culturemech_term_id, mim_id, ...])`, so **our `term.id` outranks MIM's**. When MIM corrects a mapping the consumer still picks ours; MIM can only fix rows we hold no opinion on. Before this PR the only script reading `mappings/ingredient_mappings.sssom.tsv` was `audit_missing_roles.py`, and only for role coverage — **no grounding path read it at all**.

Aligned to MIM `mapping_set_version: 2026-08-18`, which the report records, so "which MIM release did we align to" is answerable from the artifact rather than from memory.

## The picture is much better than #256 recorded

```
DIVERGENT            12 names (879 rows)
INTERNAL_SPLIT       75 names (7,888 rows)
MISSING_GROUNDING    62 names (237 rows)
```

#256's headline was "6,680 divergent rows, 115 distinct names". Both sides have curated since. Of the cases it named by hand:

| ingredient | #256 | today |
|---|---|---|
| `glucose` | CHEBI:42758 aldehydo-D-glucose | **CHEBI:17234 — agrees with MIM** |
| `magnesium sulfate heptahydrate` | CHEBI:86463 **potassium aluminium sulfate** | **CHEBI:31795 — correct** |
| `mgso4·7h2o` vs `magnesium sulfate heptahydrate` | two ids for one substance | **both CHEBI:31795 — inconsistency gone** |
| `4-aminobenzoic acid` | CHEBI:194474 | **CHEBI:30753 — the term #256 preferred** |

MIM's `aba51551 Merge the aldehydo- duplicate pairs onto the class term (#398)` is what closed the glucose case from their side.

## The measurement needed care to be worth anything

Comparing `term.id` to MIM reports **24,400 false divergences**. MediaDive-derived records keep the source's own id there and the ontology grounding in `chebi_term.id`:

```yaml
term:       {id: mediadive.compound:5, label: Glucose}
chebi_term: {id: CHEBI:17234, label: glucose}
```

Reading `chebi_term` first drops it to 3,385 real ones. Two other guards: only `skos:exactMatch` counts (a `narrowMatch` asserts a relationship, not an identity — MIM maps `Dry cow-manure` narrowMatch to `ENVO:00003031 animal manure`), and a name MIM maps two ways is skipped rather than guessed at.

## Read-only, because divergence runs both ways

- `MgCl2 x 6 H2O` — our hydrate CHEBI:86345 against MIM's anhydrous CHEBI:6636. **Ours is better.**
- `Sodium phosphate dibasic` — our CHEBI:34683 (Na₂HPO₄) against MIM's CHEBI:37583 **trisodium phosphate**. Dibasic is Na₂HPO₄, so **MIM looks wrong** and this is worth reporting back.
- `EDTA` — our CHEBI:64755 EDTA(2-) against MIM's CHEBI:4735 neutral acid as supplied. **MIM is better.**

But one is not arguable, and the report surfaced it immediately:

> **`Cholesterol` → CHEBI:140435 = `cholesterol-2,2,3,4,4,6-d6`** — a deuterated mass-spec internal standard, not cholesterol. MIM's CHEBI:16113 is plainly right.

`INTERNAL_SPLIT` is independent of MIM and often unambiguous: `dipotassium phosphate` is CHEBI:131527 on 475 rows and CHEBI:63036 on **1**.

## The gate

```
just audit-mim-sssom      # --max-divergent 12 --max-split 75
```

Baselines are **names, not rows** — one regrounding decision fixes every row of a name, so names are what a curator works through. Verified to bite when lowered by one.

## Notes

- 17 tests, concentrated on the two places a name join can lie: which slot holds our CHEBI, and which predicates count as a verdict.
- Scans all three reagent locations, so stock-solution records are not silently exempt.
- Needs MediaIngredientMech checked out beside this repo; `--sssom` overrides, and a missing file fails with a usable message rather than an empty report.
- `derived_artifacts.tsv` refreshed. 1,309 tests pass, ruff clean.

**Not done here:** the actual regrounding. #256 is explicit that this "needs curation rather than a bulk overwrite", and this PR is what makes that curation reviewable.

### DIFF
--- a/data/import_tracking/derived_artifacts.tsv
+++ b/data/import_tracking/derived_artifacts.tsv
@@ -33,6 +33,7 @@ data/import_tracking/reports/kg_fallback_phase2_changes.tsv	SNAPSHOT	one-time mi
 data/import_tracking/reports/kg_fallback_phase2_curator_review.tsv	SNAPSHOT	one-time migration (migrate_kg_fallback_phase2_oak_ols.py)		scripts/migrate_kg_fallback_phase2_oak_ols.py	
 data/import_tracking/reports/kg_fallback_phase3_changes.tsv	SNAPSHOT	one-time migration (migrate_kg_fallback_phase3_curation.py)		scripts/migrate_kg_fallback_phase3_curation.py	
 data/import_tracking/reports/mim_grounding_divergences.tsv	SNAPSHOT	one-time migration (migrate_legacy_mim_to_chebi.py)		scripts/migrate_legacy_mim_to_chebi.py	
+data/import_tracking/reports/mim_sssom_divergence.tsv	CURRENT_VIEW	audit_mim_sssom_divergence.py	scripts/audit_mim_sssom_divergence.py	scripts/audit_mim_sssom_divergence.py	
 data/import_tracking/reports/missing_compositions.tsv	CURRENT_VIEW	triage_missing_compositions.py	scripts/triage_missing_compositions.py	scripts/triage_missing_compositions.py	yes
 data/import_tracking/reports/name_term_element_mismatch.tsv	CURRENT_VIEW	audit_name_term_elements.py	scripts/audit_name_term_elements.py	scripts/audit_name_term_elements.py	
 data/import_tracking/reports/primary_term_canonicalize_changes.tsv	UNKNOWN	no writer found			
--- a/data/import_tracking/reports/mim_sssom_divergence.tsv
+++ b/data/import_tracking/reports/mim_sssom_divergence.tsv
@@ -1,0 +1,150 @@
+finding	ingredient	our_ids	our_label_asserted	mim_id	mim_label	rows	detail
+INTERNAL_SPLIT	(NH4)3 citrate	CHEBI:63037=40|CHEBI:72699=4	CHEBI:63037=triammonium citrate|CHEBI:72699=(2R)-trihomocitrate(3-)	CHEBI:63037	triammonium citrate	44	2 CHEBI ids for one name; MIM matches one of them
+INTERNAL_SPLIT	(NH4)H2PO4	CHEBI:62982=3|CHEBI:39745=1	CHEBI:62982=ammonium dihydrogen phosphate|CHEBI:39745=dihydrogenphosphate			4	2 CHEBI ids for one name; MIM has no opinion
+INTERNAL_SPLIT	(NH4)HCO3	CHEBI:184335=2|CHEBI:17544=1	CHEBI:184335=Ammonium bicarbonate|CHEBI:17544=hydrogencarbonate			3	2 CHEBI ids for one name; MIM has no opinion
+DIVERGENT	1,2-Dichloropropane	CHEBI:142468=2	CHEBI:142468=1,2-dichloropropane	CHEBI:82163	rac-1,2-dichloropropane	2	we and MIM both ground this name, to different ids
+MISSING_GROUNDING	4-Hydroxy-L-proline			CHEBI:18240	4-hydroxy-L-proline	2	MIM grounds this name and we do not
+MISSING_GROUNDING	4-Hydroxybenzaldehyde			CHEBI:17597	4-hydroxybenzaldehyde	2	MIM grounds this name and we do not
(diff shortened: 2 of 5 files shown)


realmarcin:
Adversarial review pass. One honesty defect fixed here (16b583f); two findings filed.

## Fixed — the gate did not say what it could see

The audit reported "12 divergent names" without stating what it compared them against. MIM's exactMatch labels reach **517 of our 3,519 distinct ingredient names — 14.7%**. A green run therefore reads as "we agree with MIM" when it mostly means "MIM has not looked at these". Coverage is now part of the output:

```
comparable: 517 of 3,519 names (14.7%), 125,081 of 206,344 rows (60.6%)
the rest are names MIM's exactMatch labels do not cover, so this gate cannot see them
```

The row figure is the reassuring one — MIM covers the common reagents. The name figure is the honest one.

## Checked and clean

- **No MIM name is dropped for mapping two ways** under exactMatch, so the ambiguity guard costs zero coverage today.
- **Normalization folds exactly one pair** of MIM labels together — `Methyl Beta-D-xylopyranoside` / `methyl-beta-D-xylopyranoside` — which is the fold working, not a false merge.
- **Predicate split in the SSSOM**: 2,236 exactMatch, 566 closeMatch, 142 narrowMatch. Only exactMatch is treated as a verdict, and the narrowMatch example (`Dry cow-manure` → `ENVO:00003031 animal manure`) shows why.

## Filed, not fixed

- **#304** — synonym matching would take name coverage from 14.7% to **30.2%** (+544 names, +27,845 rows), mostly closing the hydrate-spelling gap (`mgcl2 x 6 h2o` and friends). Deliberately not folded in: it moves both baselines, and landing coverage plus a gate in one step pins the ratchet to findings nobody reviewed. Suggested path is a flag, a diff of the two runs, then re-baseline in the same PR.
- **#305** — `Cholesterol` → CHEBI:140435 is `cholesterol-2,2,3,4,4,6-d6`, a deuterated mass-spec standard. Unlike the rest of the #256 backlog this is not a judgement call, so it is split out to be fixed directly.

## One thing worth flagging for the reviewer

`DIVERGENT` counts only names where we hold **one** id that differs from MIM. A name that is *both* internally split and disagrees with MIM is reported as `INTERNAL_SPLIT` (with MIM's verdict shown in the row), so the headline 12 undercounts total disagreement — the raw figure is 41 names / 3,385 rows. That is deliberate: for a split name the split is the first problem to fix, and double-counting would make the two baselines move together. No information is lost, but the 12 should not be read as "only 12 names disagree with MIM".

1,309 tests pass, ruff clean.

realmarcin:
Second adversarial pass. One gap fixed here (474bff7); one filed; three risks checked and clean.

## Fixed — bare rows on a partly-grounded name were invisible

`MISSING_GROUNDING` fired only when a name was **wholly** ungrounded. That hid 96 rows: 184 names carry some grounded rows and some bare ones, and MIM has an opinion on 48 of them. `Pyrrole-2-carboxylic acid` is grounded on 1 row and bare on 18 — the cheapest fix in the report, since MIM has already decided and most rows of the name already agree.

110 names (333 rows) now, up from 62 (237). **No baseline moves**: neither ratchet (`--max-divergent`, `--max-split`) keys on this finding, so nothing gets pinned to numbers nobody reviewed.

## Checked and clean

- **CI has no MIM checkout.** `mim_sssom_divergence.tsv` is *not* in `audit_derived_artifacts`' `CHECKABLE` set, so CI never tries to regenerate it. A missing SSSOM exits 1 with a usable message rather than writing an empty report — verified: `missing SSSOM -> exit=1`.
- **No test depends on the sibling checkout** — every `audit()` call in the suite passes a fixture SSSOM.
- **The pending NBRC recovery (#306) does not move the baselines.** Run against that branch's corpus: `DIVERGENT 12`, `INTERNAL_SPLIT 75` — both unchanged, with one extra `MISSING_GROUNDING` row. The committed report will want a refresh after that merge, but not a re-baseline.

## Filed — #307

`DIVERGENT` counts only names where we hold exactly one id that differs, so a name that is both split *and* divergent is filed as `INTERNAL_SPLIT`. Headline 12 vs raw 41 names / 3,385 rows. No information is lost — the split rows carry MIM's id and say whether it matches — but the headline shouldn't be read as "only 12 names disagree". Deliberate, because for a split name the split is the first thing to fix, and double-counting would make one fix look like two regressions.

18 tests, ruff clean, CI green on all four checks.

PR #306, opened 2026-08-19
Recover 62 NBRC composition tables from the preserved source HTML (#299)

realmarcin:
The recovery half of #299. **All 31 garbage solution nodes leave the graph, and 62 composition tables come back.**

## The raw JSON was a dead end; the HTML was not

`data/raw/nbrc/nbrc_media.json` is corrupt in *exactly the same way* as the normalized records — its `ingredients[0].name` is the same concatenated blob — so the defect is in the **scraper**, and re-importing from the JSON reproduces it.

But the scraper's own HTML captures survived, and NBRC serves the composition as a nested table with one cell per field:

```html
<td colspan='4'>MgSO<sub>4</sub>&middot;7H<sub>2</sub>O</td>
<td align='right'>0.5</td>
<td align='left'>g</td>
```

The blob is what `get_text()` produces from that with no separator. The delimiters were never in the data we normalized — they are still in the HTML. **All 62 affected records have a preserved page.**

## This resolves what #166 concluded could not be resolved

#166 tried to recover the same records by splitting the crammed **string**, and found the ambiguity is chemical: `KH2PO40.85g` splits equally well into `KH2PO` + `40.85g`, both round-trip, and one is a non-existent compound at 47× the real concentration. It therefore refused to write, correctly. Going back to the source **removes** the ambiguity instead of adjudicating it.

## Result

| | before | after |
|---|---:|---:|
| `NAME_IN_CONCENTRATION` | 59 | **0** |
| `UNPARSED_SOLUTION_TABLE` | 31 | **0** |
| total audit findings | 169 | **64** |
| garbage solution nodes in the KG | 31 | **0** |

Export delta: **−31 blob edges, +64 real edges** from the newly nested stock solutions. 0 dangling, 0 orphaned.

## Two defects the verification caught — both would have made things worse

**26 groundings destroyed.** Four records (NBRC_815, 1038, 1245, 1298) already held a curated, grounded ingredient list, and for them the HTML is a **downgrade**: NBRC_1245 carries `Calf brains`, `Beef heart`, `Proteose peptone`, `Dextrose` where the page lists only `Bacto Brain Heart Infusion (Difco)`. Wholesale replacement dropped 26 `has_part` edges to CHEBI/MICRO/FOODON/UBERON — found by diffing the export, not by inspection. Those four are now repaired **surgically**: the corrupt row and the blob go, the curated list stays. Grounding is the signal, not row count — in the other 58 the only "named" ingredient *is* the blob.

**Flattened cocktails, caught by `audit-concentration-plausibility`.** NBRC nests stock solutions as footnotes, with the medium referencing them (`Trace element solution** 2 ml`). A first version flattened those into the ingredient list and the audit gained **21 rows** — `FeCl3·6H2O 1.35 G_PER_L` among them, trace-element *stock* strength presented as a final concentration. That is precisely the #150 defect, and the gate caught me committing it. Sections are now split at their headers and nested under `solutions[]`.

NBRC_1185 needed a second pass: its header is `Trace elements:` — colon-terminated, not starred. And a header with no ingredient rows after it is a supplier note (`*Wako Pure Chemical Ind., Ltd., Osaka, Japan`) or an instruction, not a solution, so it stays prose. **Net new implausible rows: 0.**

## Consequential changes, handled rather than bulldozed

Six other tests moved because the corpus improved:

- `composition_type` / `medium_type` — NBRC_1158's real composition is now visible and contradicts its `DEFINED` claim. Restamped via `just audit-composition-type --apply` and `just curate-medium-type --apply`, i.e. the repo's own recipes.
- Indexes regenerated; derived reports refreshed via `just refresh-derived`.
- `test_unparsed_compositions` — the 25-record backlog is empty. The tool is kept as a **detector**: if an import crams a composition again, the worklist returns.
- `test_review_need` — its NBRC_1197 anchor was retired, not swapped. That record now scores 35 for the accurate, milder reason "no ingredient is grounded", and **no remaining record is both independently-confirmed-broken and top-60**, so naming a substitute would assert a claim nothing backs. The test now checks the property that matters: every top-ranked record earned it.

## Notes

- Preparation prose is preserved in `notes`, not discarded.
- Every changed record passes `validate-strict` (62/62, 0 errors).
- `data/raw/nbrc/scraped/` is gitignored, so the recovery only *runs* on a machine with the scrape — the recovered YAML is what ships.
- Dry-run by default; `just recover-nbrc-composition --limit 1` canaries one record.
- 1,292 tests pass.


realmarcin:
Adversarial review pass, run after rebasing onto current `main` so #303's audit is present.

## Fixed here — the MIM divergence report was stale

#303 landed while this branch was open, and its committed report was generated from the **pre-recovery** corpus. Restoring 62 compositions adds ~1,051 named ingredients, so it drifted. Refreshed in ee32705, and the numbers are the useful part:

```
DIVERGENT           12 names   (unchanged)
INTERNAL_SPLIT      75 names   (unchanged)
MISSING_GROUNDING  110 -> 152 names,  333 -> 823 rows
```

**Nothing new disagrees with MIM, and no name gained a second grounding** — the recovery introduces no conflict. What it does is surface **42 more names MIM has already grounded**, filed as #308. Neither ratchet moves, so `--max-divergent 12 --max-split 75` still hold.

## Checked across all 62 changed records — clean

| check | result |
|---|---|
| `notes` clobbered by the prose append | **0** |
| duplicate ingredient name within one list | **0** |
| records left with an empty ingredient list | **0** |
| runaway parse (>60 ingredients) | **0** |
| `validate-strict` errors | **0 / 62** |
| new implausible concentrations | **0** |
| dangling / orphaned KGX ids | **0 / 0** |

## Standing caveat, worth a reviewer's attention

The recovered ingredients are **ungrounded**, so they emit no KGX edges yet — `medium_to_ingredient_edge` returns `None` without a `term.id`. The graph gain in this PR is the removal of 31 garbage solution nodes plus 64 real edges from the newly nested stock solutions; turning 1,051 recovered names into graph content is #308.

## Prior passes, for the record

The two defects that mattered were caught during development, not here, and both would have made things worse: **26 groundings destroyed** on four records where the HTML is a downgrade (found by diffing the export), and **21 flattened cocktails** (caught by `audit-concentration-plausibility` — the #150 gate catching a regression in this PR).

All 6 checks green. Rebased cleanly on `main`; no conflicts.

PR #309, opened 2026-08-20
Let client flags reach deep-research-client again (#297, #289)

### LINKED ISSUE #297: research_media.py cannot forward client flags: argparse.REMAINDER rejects --max-cost and friends
Found while adversarially reviewing PR #293. **Pre-existing on `main`** — not
introduced by that PR — and verified identical under both arg shapes, so it is
filed rather than folded in.

## The defect

`scripts/research_media.py` collects extra arguments with

```python
parser.add_argument("passthrough_args", nargs=argparse.REMAINDER)
```

and `build_command` appends them to the `deep-research-client` invocation. But
`argparse.REMAINDER` on a positional does not capture a leading-`--` token that
appears before any positional: argparse claims it as an unknown *option* first.
So the documented way to pass client flags does not work:

```
$ uv run python scripts/research_media.py --provider claude_code --target ko2_no3 \
    --template templates/media_growth_research.md --research-dir research \
    --max-cost 1 --dry-run
research_media.py: error: unrecognized arguments: --max-cost
```

Identical on the `--focus` arg shape from #293, which is how I confirmed it is
not a regression:

```
$ ... --provider claude_code --target ko2_no3 --focus=growth_evidence \
    --research-dir research --max-cost 1 --dry-run
research_media.py: error: unrecognized arguments: --max-cost
```

The unit tests do not catch it because they call `build_command` directly with
`passthrough_args=["--max-cost", "1"]`, which works fine. The break is purely in
CLI argument routing.

## Why it matters

Every `just` recipe in the Research group ends with `{{args}}`, and the recipes
and skills advertise passing client options that way. In practice only
*non-flag* trailing tokens survive, so `--max-cost`, `--model`, and any other
`deep-research-client` flag are unreachable through the runner. Callers have
been silently limited to the defaults.

## Options

1. **`parse_known_args`** — take the unrecognized list as the passthrough. Simple,
   and matches what callers expect, but it silently swallows genuine typos in
   this script's own flags (`--focsu formulation` would be forwarded to the
   client rather than rejected).
2. **An explicit `--` separator** — keep `REMAINDER` but document and test
   `... --target ko2_no3 -- --max-cost 1`. Unambiguous, no typo-swallowing, but
   it changes the documented call shape.
3. **A dedicated repeatable flag** — `--client-arg --max-cost --client-arg 1`.
   Most explicit, most verbose.

I lean to 2: it keeps typo detection, and `--` is the conventional signal for
"everything after this belongs to the child process". Whichever is chosen, the
`just` recipes, `docs/RESEARCH_ARTIFACT_CONTRACT.md`, and the
`deep-research-medium` skill need updating together, plus a CLI-level test —
the existing coverage tests the function, not the parser.

## Scope

Likely affects the sibling runners too (`research_trait.py`,
`research_ingredient.py`, `research_community.py`) if they use the same
`REMAINDER` pattern — worth checking before fixing, since #289 is trying to give
these one contract.

Reproducible with `--dry-run`; no provider call and no credits.

realmarcin:
Closes #297. Part of #289's "one consistent operational contract".

## The defect

`argparse.REMAINDER` on a positional does not capture a leading-`--` token that appears before any positional — argparse claims it as an unknown option first. So the documented way to pass client options simply failed:

```
$ ... --target ko2_no3 --max-cost 1 --dry-run
research_media.py: error: unrecognized arguments: --max-cost
```

**Every `deep-research-client` flag was unreachable through the runner**, and callers were silently limited to the defaults. Every recipe in the Research group ends with `{{args}}` and the skills advertise passing client options that way, so this had been quietly broken.

The existing tests missed it because they call `build_command` directly with `passthrough_args=["--max-cost", "1"]`, where passthrough always worked. The break was purely in CLI argument routing — so the new test is at that layer.

## The fix

A `--` separator fixes the routing, but `REMAINDER` then captures the separator **too**, and forwarding a bare `--` makes it an argument of the child. Stripping the leading one is what makes the conventional shape work:

```
$ just research-media claude_code ko2_no3 --dry-run -- --max-cost 1
... --provider claude_code --output research/media/bacterial/ko2_no3-...md --max-cost 1
```

Only the **leading** separator is removed. A later `--` belongs to the child — some CLIs use it to mark their own positionals — so it is passed on.

## Why not `parse_known_args`

It would forward this script's own typos to the client instead of rejecting them. `--focsu formulation` should be an error, not a mystery flag arriving at the child.

## Notes

- Documented in the recipe comment and in `docs/RESEARCH_ARTIFACT_CONTRACT.md`.
- Verified through both entry points — direct and via `just`.
- No provider called, no credits: `--dry-run` throughout.
- 26 tests in the runner suite, ruff clean.

### DIFF
--- a/docs/RESEARCH_ARTIFACT_CONTRACT.md
+++ b/docs/RESEARCH_ARTIFACT_CONTRACT.md
@@ -33,6 +33,22 @@ non-default focus, or use `research-entity`.
 
 `just research-focuses` prints the table below from the code.
 
+### Passing options to deep-research-client
+
+Client flags go after a `--` separator:
+
+```
+just research-entity claude_code ko2_no3 formulation --dry-run -- --max-cost 1
+just research-media  claude_code ko2_no3 --dry-run -- --max-cost 1
+```
+
+Without the separator, argparse claims the flag as one of the runner's own and
+rejects it (`unrecognized arguments: --max-cost`). That is #297, and until it was
+fixed no client flag was reachable through the runner at all — the unit tests
+missed it because they call `build_command` directly, where passthrough always
+worked. The separator itself is stripped before the command is built, so the
+child never sees a bare `--`.
+
 ### Focuses
 
 | focus | template | question it asks |
--- a/project.justfile
+++ b/project.justfile
@@ -47,6 +47,10 @@ fetch-raw-data:
 # here, which is what let `--focus formulation` rank providers for formulation
 # work and then research growth evidence anyway.
 #
+# Options for deep-research-client go after a `--` separator:
+#     just research-entity claude_code ko2_no3 formulation --dry-run -- --max-cost 1
+# Without it argparse claims the flag as the runner's own and rejects it (#297).
+#
 # `focus` is POSITIONAL, so flags must come after it:
 #     just research-entity claude_code ko2_no3 formulation --dry-run
 #     just research-entity claude_code ko2_no3 growth_evidence --dry-run
--- a/scripts/research_media.py
+++ b/scripts/research_media.py
@@ -532,8 +532,34 @@ def parse_args(argv: list[str]) -> argparse.Namespace:
         action="store_true",
         help="Print the deep-research-client command without running it.",
     )
-    parser.add_argument("passthrough_args", nargs=argparse.REMAINDER)
-    return parser.parse_args(argv)
+    parser.add_argument(
+        "passthrough_args",
+        nargs=argparse.REMAINDER,
+        help="Arguments for deep-research-client, after a `--` separator: "
+             "`... --target ko2_no3 -- --max-cost 1`",
+    )
+    args = parser.parse_args(argv)
+    args.passthrough_args = strip_separator(args.passthrough_args)
+    return args
+
+
+def strip_separator(passthrough: list[str]) -> list[str]:
+    """Drop the `--` that hands the rest of the line to deep-research-client.
+
+    `argparse.REMAINDER` on a positional does not capture a leading-`--` token
+    that appears before any positional — argparse claims it as an unknown option
+    first — so the documented `--max-cost 1` was simply rejected (#297):
+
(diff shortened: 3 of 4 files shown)


realmarcin:
Adversarial review pass. No defects found in the change; recording what was probed so the merge is not on faith.

## The obvious follow-on question: do the sibling runners share the bug?

Within CultureMech, **no**. `argparse.REMAINDER` appears in exactly one script:

```
$ grep -ln 'argparse.REMAINDER' scripts/*.py
scripts/research_media.py
```

`research_media_edison.py` uses ordinary typed arguments and accepts its documented flags — verified live, `--job literature-high` renders and reports `job=LITERATURE_HIGH`, no error. So an unknown flag there is *rejected* rather than silently dropped, which is the correct failure mode and needs no change.

This does still matter for the other Mechs. Their runners were written from the same template, and the symptom is silent — callers simply get the defaults — so #289's propagation note now flags it explicitly.

## Checked and clean

- **Edge cases of the separator**: `["--"]` → `[]`, `["--", "-a", "--", "-b"]` → `["-a", "--", "-b"]` (a later `--` belongs to the child), and a passthrough with no separator is untouched.
- **The separator never reaches the client** — asserted directly, since forwarding a bare `--` would make it an argument of the child process.
- **`--list-providers` / `--provider-info` / `--list-focuses`** return before passthrough is used, so they are unaffected.
- **Nothing was dirtied by the verification**: `research/` is gitignored, and the one dry-run meta the Edison check produced was removed.
- **`research-media-codex`** is not a recipe on `main` — it lives on the unmerged #284 branch — so there is no recipe pointing at a missing script here.

## Why a CLI-level test

The existing suite calls `build_command` directly with `passthrough_args=["--max-cost", "1"]`, where passthrough always worked. That is precisely why this survived: the break was in argument *routing*, one layer up from where the tests looked. The new test drives `parse_args`.

All four checks green. No provider called, no credits — `--dry-run` throughout.

Issue #315, opened 2026-08-21
deep_research_provider.py: ranking collapses to alphabetical order when every provider's adjusted score is non-positive


realmarcin:
## Problem

`rank_stage()`'s fit calculation anchors at 0 (`max(0.0, raw[name])`) and
normalizes against the maximum raw score (`high`). When every provider's
`_score()` result is non-positive — reachable via a `provider_adjustments`
config that pushes every provider's score negative, e.g. testing a "penalize
everything" policy, or a focus with unusually punishing `cost_weight`/
`speed_weight` against every provider's actual cost/time — every `fit`
becomes `round(100 * 0.0 / high)` = 0 regardless of the underlying relative
differences, since `max(0.0, raw[name])` clamps every numerator to 0 before
the division even runs. `rank_stage()`'s final sort
(`sorted(rows, key=lambda row: (-row["fit"], row["provider"]))`) then ties on
`fit == 0` for every row and falls back to pure alphabetical order by
provider name — silently discarding the actual relative ranking a curator
would want ("least bad" vs "most bad").

An earlier attempted fix (`high = max(raw.values()); if high <= 0: high =
1.0`, landed identically in CultureMech/MediaIngredientMech/CommunityMech/
ProteinTraitsMech's copies) does **not** actually solve this: it only
replaces an exact-zero `high` (a real but much narrower case — every score
landing at precisely 0 through cancelling adjustments). When `high` is
negative rather than zero, `0.0 / high` is still `0.0` — dividing zero by
any nonzero number is always zero, positive or negative — so the collapse to
alphabetical order happens identically with or without that guard.

Not currently reachable via any shipped `conf/deep_research_provider.yaml`
in the fleet — only via a deliberately adversarial or misconfigured
`provider_adjustments`/`cost_weight`/`speed_weight` combination. Filed as a
real gap, not urgent.

## Proposed scope

This needs a design decision, not a one-line fix. Options:

1. **Min-max normalization**: `low = min(raw.values())`; `fit = round(100 *
   (raw[name] - low) / (high - low)) if high > low else 50`. Preserves
   relative ranking even when every score is negative, but changes what
   `fit` means: currently "0 = non-positive absolute score, unrelated to
   what else is ranked"; with min-max, "0 = worst in *this* ranking," even
   if that worst provider still has a strongly negative absolute score. That
   semantic change should be deliberate, documented, and probably needs a
   test asserting the new floor/ceiling meaning explicitly.
[...]

PR #317, opened 2026-08-21
Validate provider_adjustments and stage capabilities keys

realmarcin:
## Summary
Adds `provider_adjustments`/`capabilities`-key validation to this canonical copy of `scripts/deep_research_provider.py`:

- `provider_adjustments` keys are canonicalized and validated against known providers at config-load time — a typo'd or aliased key no longer silently no-ops. Also rejects two raw keys (e.g. `edison`/`falcon`) that canonicalize to the same provider, which would otherwise let the second silently overwrite the first. Also rejects an explicit `provider_adjustments: null` block, which used to skip validation entirely (guarded behind `if adjustments is not None:`) and crash later in `rank_stage`/`_score` with `AttributeError` instead of a clean `ValueError` — and, separately, rejects a non-numeric `provider_adjustments` *value* (including an explicit `null`), which crashed the same way in `_score`'s `float(adjustments.get(...))`.
- Stage `capabilities` keys are validated against the union of capabilities any `PROVIDERS` entry actually declares, mirroring the `provider_adjustments` check — a typo'd capability key previously contributed 0 to every provider's score with no error. Capability dict keys are coerced to `str` before diffing, closing a `TypeError` when YAML 1.1 parses an unquoted key like `no:` as bool `False`. Capability weight *values* and the stage-level `synthesis_weight`/`speed_weight`/`cost_weight` scalars are validated numeric too (including rejecting a YAML boolean, since `bool` subclasses `int`), for the same crash-vs-clean-error reason.

**Provenance correction** (an earlier revision of this description got this backwards): this PR is the *originating* fix, not a port. ProteinTraitsMech#487, MediaIngredientMech#412, and CommunityMech#657 (the spoke PRs originally cited as the source) do **not** contain any of this validation — none of them, verified directly. The three spoke ports that actually carry this logic (ProteinTraitsMech#514, MediaIngredientMech#418, CommunityMech#662) are downstream of *this* PR, still open pending it, and each states so in its own description. CultureMech#328 tracks that dependency. Flagged by dynamic-review.

**Beyond the validation work above**, review while backporting proteintraitsmech#514's fixes into this same branch surfaced several more real, independent gaps that are now also fixed here:
- `--allow ""` (an explicit empty string) used to be indistinguishable from omitting the flag, silently disabling all `--allow` filtering — a real policy-bypass bug, not a doc nit. Now rejected with the same "did not contain any provider names" error as other empty-after-strip `--allow` values.
- `--allow mock` passed validation (mock is a real `PROVIDERS` key) but `recommendable()` unconditionally excludes it — structurally impossible to recommend regardless of `--allow`. Now rejected explicitly.
- `print_report`'s fallback message was wrong when `--allow`/`--no-paid` policy, not missing credentials, emptied a stage's recommendation; it now names the actual cause and routes through the single `recommendable()` helper rather than re-implementing its filter as a second, independently-maintained condition.
- `test_a_measured_dead_provider_is_not_recommended`'s end-to-end assertion was vacuous under real CI conditions (no credential set at all, and even after fixing that, a `None`-branch escape hatch let it pass regardless of whether the block was actually enforced) — reproduced live, fixed to assert against `recommendable()` directly.

Also documents (does **not** fully fix) a real, separate gap found while porting: `high = max(raw.values()) or 1.0` only guards an exact-zero max. When every provider's adjusted score is negative, `fit` is `0/high` either way (0.0 divided by any nonzero number, positive or negative, is still 0.0), so the ranking still collapses to alphabetical order. Fixing that needs a real normalization design decision (min-max vs. the current absolute-zero-floor semantics), not a one-line change — filed as #315. I initially wrote a "fix" that didn't actually address this (confirmed via test failure, not just review) and corrected course rather than merge something that only looked fixed. Note: `test_exact_zero_max_score_does_not_divide_by_zero` covers only the pre-existing exact-zero case, not the new `if high <= 0` branch's negative-`high` path specifically (both clamp to the same result for that test's all-zero input) — a genuine but low-priority test-precision gap, not a functional one.

## Test plan
- [x] Full `uv run --extra dev pytest tests/` — 1117 passed, 17 skipped, 0 failed (re-verified against current head; an earlier revision of this description cited a stale count from an earlier commit)
- [x] `ruff check` on changed files — clean
- [x] Mutation-tested every new fix (capabilities-key validation, `--allow ""`/`mock` rejection, str-coercion, PAID_COSTS medium-cost blind spot, `provider_adjustments`/capability/weight null and bool values, vacuous-assertion fix) — each reverts to a red test
- [x] Confirmed the real config (`conf/deep_research_provider.yaml`) still loads cleanly

### DIFF
--- a/scripts/deep_research_provider.py
+++ b/scripts/deep_research_provider.py
@@ -228,6 +228,9 @@ PROVIDERS: dict[str, Provider] = {
 }
 
 ALIASES = {"edison": "falcon", "futurehouse": "falcon", "claude-code": "claude_code"}
+_ALL_CAPABILITIES = frozenset(
+    cap for provider in PROVIDERS.values() for cap in provider.capabilities
+)
 COST_VALUE = {"low": 1, "medium": 2, "high": 3, "very_high": 4}
 TIME_VALUE = {"fast": 1, "medium": 2, "slow": 3, "very_slow": 4}
 SYNTHESIS_VALUE = {"none": 0, "summary": 1, "deep": 2, "agentic": 3}
@@ -345,6 +348,36 @@ def load_config(path: Path) -> dict[str, Any]:
                 raise ValueError(
                     f"Stage {focus_name}.{stage_name}.capabilities must be a mapping"
                 )
+            unknown_caps = set(capabilities) - _ALL_CAPABILITIES
+            if unknown_caps:
+                raise ValueError(
+                    f"Stage {focus_name}.{stage_name}.capabilities names unknown "
+                    f"capability/ies {sorted(unknown_caps)}; no provider declares "
+                    f"them, so they would silently score 0"
+                )
+        adjustments = focus.get("provider_adjustments")
+        if adjustments is not None:
+            if not isinstance(adjustments, dict):
+                raise ValueError(
+                    f"Focus {focus_name!r}.provider_adjustments must be a mapping"
+                )
+            canonical: dict[str, Any] = {}
+            for raw_name, value in adjustments.items():
+                name = canonical_provider(str(raw_name))
+                if name not in PROVIDERS:
+                    raise ValueError(
+                        f"Focus {focus_name!r}.provider_adjustments names unknown "
+                        f"provider {raw_name!r} (resolved to {name!r}); "
+                        f"known providers: {', '.join(sorted(PROVIDERS))}"
+                    )
+                if name in canonical:
+                    raise ValueError(
+                        f"Focus {focus_name!r}.provider_adjustments has multiple "
+                        f"keys resolving to provider {name!r} (e.g. {raw_name!r}); "
+                        f"use a single canonical key per provider"
+                    )
+                canonical[name] = value
+            focus["provider_adjustments"] = canonical
     return data
 
 
@@ -376,7 +409,18 @@ def rank_stage(
         name: _score(provider, stage, adjustments)
         for name, provider in PROVIDERS.items()
     }
-    high = max(raw.values()) or 1.0
+    # `or 1.0` only replaces an exact-zero max (e.g. every raw score landing
+    # at precisely 0 through cancelling adjustments) — a real, if rare, case
+    # this guards. It does NOT fix the harder case where every score is
+    # negative: fit is 0/high either way there (0.0 divided by any nonzero
+    # number is 0.0), so the ranking still collapses to alphabetical order
+    # once every candidate is "actively bad" rather than merely "not the
+    # best." That needs a different normalization (e.g. min-max instead of
+    # max-only) and is a real design decision, not a one-line fix — see
+    # CultureMech#315.
+    high = max(raw.values())
+    if high <= 0:
+        high = 1.0
     rows = []
     for name, provider in PROVIDERS.items():
         status, reason = provider_status(name)
--- a/scripts/deep_research_provider.py
+++ b/scripts/deep_research_provider.py
@@ -548,8 +548,7 @@ def print_report(report: Mapping[str, Any], provider_name: str | None = None) ->
             if fallback:
                 message += f"; cross-check/fallback: {fallback['provider']}"
             print(message)
-        elif any(row["status"] == "available" and row["provider"] != "mock"
-                 for row in stage["ranking"]):
+        elif recommendable(stage["ranking"]):
             print(
                 "Route now: no provider passes the current --allow/--no-paid "
                 "filters, though at least one is otherwise available."
--- a/scripts/deep_research_provider.py
+++ b/scripts/deep_research_provider.py
@@ -548,6 +548,12 @@ def print_report(report: Mapping[str, Any], provider_name: str | None = None) ->
             if fallback:
                 message += f"; cross-check/fallback: {fallback['provider']}"
             print(message)
+        elif any(row["status"] == "available" and row["provider"] != "mock"
+                 for row in stage["ranking"]):
+            print(
+                "Route now: no provider passes the current --allow/--no-paid "
+                "filters, though at least one is otherwise available."
+            )
         else:
             print(
                 "Route now: no real provider is currently available; configure a listed credential or CLI."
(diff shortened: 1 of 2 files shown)


realmarcin:
Addressing the review:

- **Low** (unknown-capability error should list known-good values): fixed in c7c0a258df.
- **Medium** (PR description implies fleet parity that doesn't exist yet): correct as written at review time — ProteinTraitsMech, MediaIngredientMech, and CommunityMech did not yet have this validation when this PR was opened. All three are being brought to parity with this exact change in the same work session (companion PRs incoming); proteintraitsmech#505 (the duplicate-alias tracking issue) will close via that port rather than staying open as a separate follow-up.

realmarcin:
Round-3 fix pushed, backporting two real bugs found while porting this file's feature set to proteintraitsmech (dynamic-review on proteintraitsmech#514):

- \`--allow ""\` (an explicitly empty string) was indistinguishable from omitting the flag entirely, because Python treats \`""\` as falsy — bypassing the "did not contain any provider names" check and silently recommending with no filtering. Fixed by checking \`args.allow is not None\` instead of \`args.allow\`. Mutation-verified.
- \`print_report\`'s fallback message ("no real provider is currently available; configure a listed credential or CLI") was wrong when \`--allow\`/\`--no-paid\`, not missing credentials, emptied the recommendation. Added a distinct message for that case.

Also added a mutation-resistant \`--no-paid\` regression test — the existing one could pass even with \`no_paid\` filtering fully removed, since the ambient test env never puts a genuinely paid provider in the winning position.

These two fixes were previously added only to proteintraitsmech's vendored copy with no tracking issue for backporting — this is that backport, landing in the hub first as intended.

Full \`uv run --extra dev pytest tests/\` — 1108 passed, 17 skipped.

realmarcin:
Round-4 fix pushed for the real finding: coerced capability dict keys to \`str\` before diffing against \`_ALL_CAPABILITIES\`, matching the existing coercion pattern used for \`provider_adjustments\` a few lines below. PyYAML's \`safe_load\` parses an unquoted \`no:\` key as the Python bool \`False\`; mixed with a genuine unknown string capability, \`sorted(unknown_caps)\` raised \`TypeError\` (str vs bool) instead of the intended clean \`ValueError\`. Reproduced live, mutation-verified.

On the second finding (missing \`Co-Authored-By\` trailer on the "Reject \`--allow \"\"\` bypass" commit): confirmed, and not amending that pushed commit to add it retroactively (avoiding a history rewrite mid-review). Noting it here as an accepted gap for that one commit.

realmarcin:
Round-5 fix pushed for the medium finding: \`test_policy_filtered_empty_recommendation_names_the_actual_cause\` makes an exact assertion on the printed message, unlike its permissive siblings (which use \`recommended is None or recommended[...] == X\` specifically to tolerate ambient credentials). An ambient \`OPENAI_API_KEY\` makes the \`--allow\`'d \`openai\` itself "available", routing through the pre-existing recommended-provider branch instead of the one under test. Fixed with \`monkeypatch.delenv("OPENAI_API_KEY", raising=False)\`. Reproduced live before the fix (\`env -i ... OPENAI_API_KEY=sk-test-ambient\` failed), confirmed passing after.

On the static-gate mismatch noted in the review (dispatched against culturebotai-claw instead of CultureMech): that's a review-infrastructure issue in this run, not a code problem — no action needed on this PR.

realmarcin:
Round-6 fix pushed for the two actionable low findings:
- Aligned "Known capabilities: ..." and "known providers: ..." with the file's one established phrasing for this error shape, "choose one of: ...".
- Filed #328 to track syncing this PR's hub-only fixes to the 3 spoke repos (noting that proteintraitsmech#514 and MediaIngredientMech#418 already picked these up in this same session; CommunityMech#662 should be checked).

Also, separately from this PR: proteintraitsmech#514's own review caught two more real gaps in the port (print_report's elif re-implementing recommendable()'s filter instead of calling it; a PAID_COSTS test blind spot for "medium"). Fixed both here in the hub first (this PR, latest 2 commits) before propagating.

realmarcin:
Round-7 fix pushed: \`provider_adjustments\` values were never validated as numeric — a profile with \`provider_adjustments: {asta: null}\` loaded cleanly and crashed later in \`_score\` with an uncaught \`TypeError\` instead of a clean \`ValueError\`, exactly the gap this PR's other fixes closed for capability weights and stage-level weight scalars. Also added bool-value coverage for all three validated fields, since the existing bool-exclusion clauses (\`isinstance(x, (int, float))\` alone would silently accept a YAML boolean, since \`bool\` subclasses \`int\`) had zero test coverage for values, only keys.

Both mutation-verified. Full \`uv run --extra dev pytest tests/\` — 1117 passed, 17 skipped.

realmarcin:
Round-8 fix pushed for a real, previously-uncaught vacuous-assertion issue flagged by dynamic-review on proteintraitsmech#514 (which ported this test): \`test_a_measured_dead_provider_is_not_recommended\` never actually set a credential for the \`build_report()\` call, and even after adding one (\`EDISON_API_KEY\`), the "\`recommended is None or ... not in KNOWN_BLOCKED\`" assertion still passed vacuously via the \`None\` branch whenever nothing else happened to be available either — reproduced live against this repo's own real conditions. Replaced with an assertion against \`recommendable(stage["ranking"])\` directly, which has no \`None\`-branch escape hatch.

Full \`uv run --extra dev pytest tests/\` — 1117 passed, 17 skipped.

realmarcin:
Round-9 fixes pushed for all 3 findings:
- Corrected the PR description's provenance claim (this PR is the *originating* fix, not a port — verified none of the three originally-cited spoke PRs contain this validation) and updated the stale test-plan count (1108 → 1117, now 1118 after this round).
- Added a non-uniform negative-score test alongside the exact-zero one. Mutation-tested it and confirmed it does **not** (and provably cannot) discriminate the guard's exact threshold: \`fit = round(100 * max(0.0, raw[name]) / high)\` clamps any negative raw score to 0.0 in the numerator before the division runs, so the divisor's value is irrelevant once scores go negative — confirmed by narrowing the guard to \`if high == 0\` and finding the new test still passes. Documented that honestly in the docstring rather than claim discrimination that doesn't exist.

Full \`uv run --extra dev pytest tests/\` — 1118 passed, 17 skipped.

realmarcin:
Updated #328 to record the three later commits and confirm current status: all of them are already propagated to and reviewed in proteintraitsmech#514, MediaIngredientMech#418, and CommunityMech#662 (same session). No code change needed on this PR — the tracking gap was in #328's own text, now corrected.

PR #333, opened 2026-08-22
Emit each KGX edge once (#312)

### LINKED ISSUE #312: KGX export emits 45,464 duplicate edge rows: shared solutions re-emit their composition per referencing medium
Found while measuring the edge gain from #308. **Pre-existing**, not introduced by
that work.

## The defect

`culturemech_edges.tsv` carries 45,464 duplicate rows — roughly 23% of the file:

```
edge rows          : 198,836
distinct edge ids  : 153,372
ids appearing >1   :  32,282   (45,464 surplus rows)
```

Every collision is an **exact duplicate triple**, confirmed:
`distinct triples == distinct ids`. So no information is lost or contradicted —
but a KGX edges file with repeated `id` values is malformed, and any consumer
that counts edges, aggregates by predicate, or loads into a store with an id
uniqueness constraint sees the wrong thing.

## Where they come from

`_make_edge_id` is a deterministic UUID5 over `subject|predicate|object`, which is
correct. The duplication is upstream of it: `transform()` walks each record's
`solutions[]` and emits `solution_to_ingredient_edge` for every entry, so a stock
solution shared across many media re-emits its whole composition once per
referencing medium:

```
x178  ('culturemech:solution_Seven_vitamins_solution', 'biolink:has_part', 'CHEBI:132751')
x178  ('culturemech:solution_Seven_vitamins_solution', 'biolink:has_part', 'CHEBI:15940')
```

`Seven vitamins solution` is referenced by 178 media, and its composition is
emitted 178 times. The medium→solution edge is genuinely per-medium and correct;
the solution→ingredient edges are a property of the solution alone.

## Fix

The same shape as the node deduplication added in #294: a run-scoped `seen` set,
keyed on the edge id, in the koza wrapper. Nodes already do this —
`_EMITTED_NODE_IDS` — and edges were simply never given the same treatment.

Worth pairing with a test that the edges file has no repeated id, alongside the
existing "no dangling `culturemech:` id" check in
`tests/test_kgx_tsv_export.py`.

## Scale note

Deduplicating removes ~23% of the rows without losing an edge. That is worth
knowing before anyone benchmarks or sizes downstream ingest against the current
file.

Verify with `just kgx-export`; read-only, no provider call.

realmarcin:
Closes #312.

## The defect

The edges file carried **45,464 duplicate rows — 23% of it**:

```
edge rows          198,836
distinct edge ids  153,372
ids appearing >1    32,282
```

`_make_edge_id` is a deterministic UUID5 over `subject|predicate|object`, which is correct. The duplication is upstream: `transform()` walks each record's `solutions[]` and emits the solution's whole composition, so a **shared stock solution re-emits it once per referencing medium**. `Seven vitamins solution` is referenced by 178 media, so each of its `has_part` edges appeared 178 times.

The medium→solution edge *is* per-medium and is correctly repeated. The solution→ingredient edges are a property of the solution alone.

## The fix

The same run-scoped `seen` set that nodes have had since #294 — edges were simply never given it.

## Nothing is lost

Every collision was an **exact duplicate triple**, confirmed by `distinct triples == distinct ids`. On the full corpus the distinct edge count is **153,372 before and after**:

```
after: 153,372 rows == 153,372 ids == 153,372 triples
       0 dangling, 0 orphaned
```

The only id-level differences between the two runs are the 3 cholesterol regroundings from #316, which landed in between — verified by diffing the two exports, not assumed.

| predicate | edges |
|---|---:|
| `biolink:has_part` | 117,701 |
| `biolink:has_attribute` | 25,147 |
| `biolink:same_as` | 10,399 |
| `biolink:subclass_of` | 48 |
| `METPO:2000517` | 77 |

## On the tests

Two, **both verified to fail when the dedup is removed**. That check earned its keep: the first version asserted no-duplicate-ids against the standard fixture, whose two records share no solution — so no duplicate could arise and it **passed with the fix reverted**. It was a vacuous test with a docstring claiming #312 coverage.

Both now use a fixture where two media reference one solution — the actual #312 shape — and assert both halves: the composition is emitted **once**, and the per-medium medium→solution edge is still emitted **twice**.

## Notes

- 1,424 tests pass.
- The 55 ruff findings in the touched files are pre-existing and identical before and after.
- Worth knowing downstream: the edges file shrinks by 23%, so any prior benchmark or ingest sizing against it was measuring inflated input.

### DIFF
--- a/src/culturemech/export/kgx_export.py
+++ b/src/culturemech/export/kgx_export.py
@@ -49,12 +49,14 @@ Legacy Edges (for backward compatibility):
 """
 
 import uuid
+from collections.abc import Iterator
 from dataclasses import asdict, dataclass
-from typing import Any, Dict, Iterator, List, Optional
+from typing import Any
 
 try:
     import koza
     from koza import KozaTransform
+
     KOZA_AVAILABLE = True
 except ImportError:
     KOZA_AVAILABLE = False
@@ -80,7 +82,8 @@ HAS_SOLUTION_COMPONENT = "biolink:has_part"  # For medium→solution (also uses 
 # PURE TRANSFORM FUNCTION (testable without Koza)
 # ================================================================
 
-def transform(record: Dict[str, Any]) -> Iterator[Dict[str, Any]]:
+
+def transform(record: dict[str, Any]) -> Iterator[dict[str, Any]]:
     """
     Pure transform function - testable without Koza.
 
@@ -216,7 +219,7 @@ class Node:
     """
 
     id: str
-    category: List[str]
+    category: list[str]
     name: str
     provided_by: str = KNOWLEDGE_SOURCE
 
@@ -260,16 +263,16 @@ class Edge:
     primary_knowledge_source: str = KNOWLEDGE_SOURCE
     knowledge_level: str = "knowledge_assertion"
     agent_type: str = "manual_validation_of_automated_agent"
-    publications: Optional[List[str]] = None
-    concentration: Optional[str] = None
-    role: Optional[str] = None
-    strain: Optional[str] = None
-    growth_phase: Optional[str] = None
-    attribute_type: Optional[str] = None
-    relationship_type: Optional[str] = None
+    publications: list[str] | None = None
+    concentration: str | None = None
+    role: str | None = None
+    strain: str | None = None
+    growth_phase: str | None = None
+    attribute_type: str | None = None
+    relationship_type: str | None = None
 
 
-def to_edge(edge_dict: Dict[str, Any]) -> Edge:
+def to_edge(edge_dict: dict[str, Any]) -> Edge:
     """Turn one ``transform`` dict into a writable KGX edge row.
 
     Unknown qualifier types are dropped rather than silently mangled into an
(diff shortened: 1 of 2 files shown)


realmarcin:
CI caught something worth recording: **`Changed Python quality + wheel smoke`** is a new gate since my last PR, and it runs Ruff **and** Black over every changed file.

Both files this PR touches already carried the debt — **55 Ruff findings and Black reformatting, none of it from the dedup**. The gate's design is "you touch it, you clean it", so the cleanup is now in `cc4e91f`, deliberately kept as **its own commit** so the four-line functional change stays reviewable against a 180-line reformat.

Almost all of it is mechanical: `UP006`/`UP045` modernising `Dict`/`Optional` to `dict`/`| None` (49 of 55), import sorting, Black's line wrapping.

**One was a real weakness in a test I wrote in #295:**

```python
with pytest.raises(Exception):        # B017
```

That passes on a typo in the field names just as happily as on the qualifier rejection it exists to prove. Now `pytest.raises(ValidationError, match="qualifiers")`, which fails if the rejection stops being about qualifiers. Ruff found a genuine gap, not just formatting.

Verified after the reformat:

```
check_changed_python.py -> All checks passed; 2 files unchanged by Black
1,424 tests pass
corpus export byte-identical to the pre-format run (153,372 edges)
```

Issue #342, opened 2026-08-26
Remediate the 2026-08-24 media-content review findings


realmarcin:
## Problem

The 2026-08-24 whole-corpus review identified blocking recipe gaps, malformed composition rows, composition-type conflicts, concentration defects, and high-frequency ingredient identity disagreements.

## Acceptance criteria

- Repair only source-backed compositions and concentrations; do not infer chemistry.
- Clear unnamed and unparsed/prose component defects.
- Resolve composition-type contradictions.
- Reconcile defensible CHEBI identities against the MediaIngredientMech SSSOM and actual ChEBI content.
- Regenerate review artifacts and lower ratchets.
- Preserve explicit worklists for source-blocked or ambiguous residuals.
- Pass affected tests, closed LinkML validation, lint, and derived-artifact freshness.

## Baseline

- 224 media missing composition
- 32 unnamed components
- 64 unparsed composition findings
- 46 composition-type conflicts
- 9,757 concentration findings / 186 flattened cocktails
- 44 exact-name grounding splits covering 3,932 rows

The implementation and final measured outcome will land from branch codex/media-content-remediation-2026-08-25. Related existing trackers: #150, #256, and #273.

realmarcin:
## Review loop status

Priority order for this issue set is #342/#273 first, then the concentration residual in #344, followed by source/identity curation in #343, #150, #256, and #151.

**Round 1:** the affected tests exposed stale post-remediation guards. That finding is tracked in #345 and addressed on the branch: concentration/cocktail ratchets were refreshed, DSMZ cobalt fixtures now use the reconciled identity, and the sodium glycerophosphate pentahydrate exception is explicit and bounded.

**Round 2:** clean. Verification completed with 414 affected tests passing, strict closed LinkML validation across 15,877 YAML records with zero errors, scoped Ruff passing, and all 9 freshness-checkable generated artifacts byte-matching regeneration.

Measured outcomes: missing compositions 224 -> 150; unnamed components 32 -> 0; unparsed composition 64 -> 0; composition-type conflicts 46 -> 0; flagged concentrations 9,757 -> 9,688; cocktail patterns 186 -> 183; 10,874 serialized wrong-ID references corrected plus 187 missing exact MIM groundings restored.

realmarcin:
## Review round 4

[...]

PR #400, opened 2026-09-02
Add single-record YAML curation skill

realmarcin:
## Summary
- add a CultureMech-specific single-record YAML curation skill and checklist
- route the skill from CLAUDE.md
- accept standard nested skill metadata while preserving the legacy layout

## Validation
- skill-creator quick validation
- targeted repository tests: 136 passed

### DIFF
--- a/.claude/skills/curate-yaml-record/SKILL.md
+++ b/.claude/skills/curate-yaml-record/SKILL.md
@@ -1,0 +1,143 @@
+---
+name: curate-yaml-record
+description: Review and curate one CultureMech medium or stock-solution YAML record for source identity, formulation accuracy, preparation detail, growth evidence, completeness, and resolvable gaps. Use when asked to audit, improve, complete, correct, or add evidence to one record; do not use for bulk ingestion, generated merge/page edits, or as permission to contact anyone or mutate GitHub.
+allowed-tools: Bash, Read, Grep, Glob, WebSearch, WebFetch, Edit, Write
+metadata:
+  category: curation
+  requires_database: false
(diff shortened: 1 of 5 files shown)


realmarcin:
## Adversarial review

One of seven copies of a single change (`feat/curate-yaml-record-skill`), reviewed as a batch — the shared content once, the per-Mech parts against this repository.

**Every guarded-write function this copy names imports.** Verified with `importlib` on the merged tree, not by reading. That is the part of a curation skill most likely to be wrong and it is right in all seven.

**It avoids the two failure modes the fleet keeps paying for.** No claim about a sibling repository (every Mech-name mention is a self-mention), and no state snapshot — no record counts, no "as of" tables. Those are exactly what culturebotai-claw#338 and #342 found rotting in the sibling skills, and this batch is the counterexample.

**Merges clean onto current `main`** (`git merge-tree`), despite the branch being behind it; the change is purely additive.

**One finding across the batch, not blocking.** In the CultureMech copy, a boundary rule names `data/raw_yaml/`, a documented data layer that is neither tracked nor gitignored — the one state no checker can classify — and the validator shipped beside it reads the filesystem rather than git, so it structurally cannot notice (CultureBotAI/CultureMech#419).

**The seven copies share a verbatim skeleton** — Boundaries, Read-before-judging, and workflow steps 1/4/6 — with steps 2/3/5 adapted per Mech. That is the managed-region shape the canonical templates use, filed as culturebotai-claw#346 so the next unification pass starts from the measured skeleton.

Merging.

PR #410, opened 2026-09-03
Advance the claw pin to the ref that declares the whole fleet

realmarcin:
**#279 step 3 of 3.**

AntibioticMech joined the fleet as a vendored consumer (CultureBotAI/AntibioticMech#177),
advancing claw's canonical ref to `0a3252d4` and leaving this repository — and five
siblings — pinned at `42b13ce3`, an ancestor of it. claw's fleet audit reads that correctly
as a rollout in progress rather than drift, and stays red until every consumer sits on one
ref:

> `0a3252d4` leads and every other pin is an ancestor of it. Advance these to `0a3252d4`:
> cellstructuremech, communitymech, culturemech, mediaingredientmech, proteintraitsmech,
> traitmech

## Content-free by construction

claw's governed artifacts are byte-identical between the two refs, so this advances the pin
and nothing else. Verified in claw **before** applying:

```
$ git diff --name-only 0a3252d4cf..HEAD -- src/kg_microbe_governance/artifacts/ \
      src/kg_microbe_governance/vendored_artifacts.json
(empty)
```

## How it was applied

Through `kg-microbe-governance sync`, not by hand — dry-run first, under the repository
lock, target resolved through `RepositorySettings` with its GitHub origin confirmed:

```
dry-run:  WOULD_WRITE  scripts/.vendored_canon_ref  content drift
apply:    WRITE        scripts/.vendored_canon_ref  content drift
```

The dry run named exactly one write and the apply made exactly that one.

MediaIngredientMech#530 was the canary for this batch: same tool, same lock, same
verification. It merged with `vendored-sync` green — the check that re-derives the artifact
set from claw's manifest *at the newly pinned commit* and compares bytes, which is what
rules out a pin moving while the content it names does not follow.

https://claude.ai/code/session_01W5NEPR2Lkxcq5Xs6nrsmCP


realmarcin:
## Adversarial review

The failure mode in a pin bump is the pin moving while the content it names does not follow, leaving the repository claiming a ref it does not carry.

**`vendored-sync` passing is exactly what rules that out** — it re-derives the artifact set from claw's manifest *at the newly pinned commit* and compares bytes. Green here, so this repository genuinely carries what `0a3252d4` declares.

**The diff is one file**, and the dry run named that file before the apply touched anything. Independently confirmed content-free: in claw, `git diff 0a3252d4..HEAD` over the artifacts directory and the manifest is empty, so the two refs declare identical bytes.

Canaried on MediaIngredientMech#530 first — same tool, same lock, same verification — which merged green before this batch was applied.

No findings. Merging.

Issue #369, opened 2026-08-28
Dependency changes cannot trigger any corpus gate: pyproject.toml and uv.lock are in no workflow's paths filter


realmarcin:
None of CultureMech's four gate workflows watch `pyproject.toml` or `uv.lock`:

```bash
for w in validate-strict chebi-consistency concentration-plausibility label-correspondence; do
  n=$(sed -n '/paths:/,/^[a-z]/p' .github/workflows/$w.yaml | grep -cE "pyproject|uv\.lock")
  echo "  $w: $n"
done
#   validate-strict: 0
#   chebi-consistency: 0
#   concentration-plausibility: 0
#   label-correspondence: 0
```

So a pure dependency change cannot trigger the gates that a dependency change is
most likely to break. `label-correspondence` is the sharpest case: it runs the
OAK id↔label check and caches `~/.data/oaklib`, so it is the gate that would
notice an oaklib regression — and it is the one a lockfile edit cannot start.

## This nearly bit us in #365

#365 bumped oaklib and it looked well covered — 7 checks, all green. But the
gates ran only because the linkml upgrade forced regenerated schema artifacts,
and `src/culturemech/schema/**` *is* in the filters. Had the scope stayed
"oaklib pin only" as originally intended, the same PR would have run `tests.yaml`
(no path filter) and nothing else.

That is a coincidence standing in for a gate.

## The fleet already found this twice

- MediaIngredientMech#495 — "Dependency bumps cannot trigger the oaklib-heavy CI
  gates: uv.lock and pyproject.toml are outside every path filter"
- TraitMech#566 — "label-correspondence's paths filter excludes pyproject.toml
  and uv.lock, so the blocking OAK gate never runs on a dependency change"

Filing here for parity; the fix is presumably the same in all three.

## Fix

Add `pyproject.toml` and `uv.lock` to the `paths:` list of each gate workflow.
Cheap, and it makes the dependency-change path exercise the same gates as a
corpus change.

Worth checking at the same time whether this interacts with #277 (path filters
see only the first 300 changed files) — a dependency bump is small, so the 300
limit is not the binding constraint here, but the two share a mechanism.

Found while extending the #362 oaklib sweep to the sibling repos.

realmarcin:
MediaIngredientMech has already fixed this, so there is a reference implementation to copy rather than a design to invent.

`.github/workflows/label-correspondence.yaml` on MIM's `main` lists both files in its `paths:`:

```yaml
[...]

PR #436, opened 2026-09-08
Re-pin claw to eeccfebe and re-vendor the id-label validator

realmarcin:
Closes nothing here; fixes culturebotai-claw#375 for this repository.

culturebotai-claw#369 added 85 lines to `scripts/validate_id_label_correspondence.py`, which `kg-microbe-governance list` shows as a governed artifact vendored into all eight Mechs. The consumers were not re-pinned, so `fleet-audit` went red on claw's `main` and reported the same row for every Mech:

```
FAIL  <mech>  canonical_ref  Pinned claw manifest does not match the installed governance package
```

It was green at `cc47ada5` and red from `e8c6af24` (#369) onward.

This carries the new artifact bytes and moves `scripts/.vendored_canon_ref` to `eeccfebe`, claw's current `main`.

## How it was produced

`kg-microbe-governance sync --apply` against a clean worktree off `origin/main`, under this repository's lock. Exactly two files, the same two in all eight Mechs.

## Verified here

- `scripts/check_vendored_sync.sh` passes: every governed artifact matches the pinned revision.
- The pin equals claw's `main` tip.
- Nothing else in the worktree changed.

The fleet-level `fleet-audit` checks committed `origin/main` tips, so it can only go green once all eight of these land; that is the point of the change rather than a gap in it.

https://claude.ai/code/session_01GhuGBogKrsE76ejRJJ2292

### DIFF
--- a/scripts/.vendored_canon_ref
+++ b/scripts/.vendored_canon_ref
@@ -1,1 +1,1 @@
-69c3abb70c4bc2671df0ccb496dc1fa374cd6dc4
+eeccfebe2369dec0f5575aff25e0350e04ec19ec
--- a/scripts/check_changed_python.py
+++ b/scripts/check_changed_python.py
@@ -11,10 +11,30 @@ from collections.abc import Sequence
 from pathlib import Path
 
 ROOT = Path(__file__).resolve().parents[1]
-EXCLUDED = {
+# Generated files, and every Python file vendored from claw. A vendored file
+# is byte-identical to claw's canonical copy by contract -- check_vendored_sync
+# fails if it is not -- so it cannot be reformatted or lint-fixed here. Only
+# check_vendored_sync.py was listed, which held only while no other vendored
+# file changed; re-vendoring validate_id_label_correspondence.py made this
+# gate and the vendored-sync gate demand opposite things of the same bytes.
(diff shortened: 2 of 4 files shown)


realmarcin:
The `Changed Python quality + wheel smoke` failure was this repository's two gates contradicting each other, not the re-pin being wrong.

`check_vendored_sync` requires `scripts/validate_id_label_correspondence.py` to be byte-identical to claw's canonical copy; `check_changed_python` requires a changed Python file to be ruff-clean and black-formatted under this repository's configuration. The vendored file carries three findings here (`UP035` and two `E741`) and must not be edited, so the two demands cannot both be met.

`EXCLUDED` already named `scripts/check_vendored_sync.py`, so the intent was there, but as a single hand-written path that held only while no other vendored file changed. Seven vendored Python files exist here; six were unlisted and each was a latent instance of the same failure. All are listed now, and #437 tracks deriving the set from claw's manifest instead of restating it.

The accompanying test earned its place on the second attempt. The first version iterated the gate's own constant, so deleting an entry shrank the assertion with it and the test passed either way; mutation caught it. It now asserts against its own literal, and both mutations go red.

Not changed: the vendored file's lint findings. They belong to claw under claw's configuration, and fixing them here would break the byte-identity the other gate enforces.

